PR #16, opened 2014-05-08
Add HOST in postgresql connection in template

tobiasBora:
Hello,

It is a minor modification but without it the template doesn't work. For the moment I'm using a similar way to register passwords (thank you Kakadu) but I'm not sure it is the better way to do it (the password is written in clear during the compilation):

```
USER := $(USER)
PASSWORD := password

# DO NOT EDIT
ENV_PGOCAML := PGHOST=$(HOST) PGDATABASE=$(DATABASE) PGPORT=$(PORT) PGUSER=$(USER) PGPASSWORD=$(PASSWORD)

```

That's why I didn't put it this pull request. If it's not the good way to configure password, could you please tell me how to do it ?

Thank you,

Tobias.


dannywillems:
I think we can close this PR because HOST is now added.

PR #30, opened 2015-10-06
README.md: add tyxml in the list of projects to pin

yetanotherion:


vasilisp:
Thank you. Merging. The need for pinning all that stuff will go away soon, with a new round of releases.

PR #31, opened 2015-10-06
README.md: add dependency to dbm

yetanotherion:
Without installing that package, ocsigenserver
is built without ocsipersist-dbm.* files.
And in that situation, eliom-base-app
fails as

$ make test.byte
ocsigenserver: ocsipersist:dbm: ...Initialization complete
ocsigenserver: ocsigen:main: Fatal - Findlib package\
eliom.server not found: maybe you forgot <findlib path="..."/>?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ If not, start with a simpler template (but you can still use the modules from El
 
 ###<a id="install"></a>Installation
 
-Eliom-base-app depends on dev version of Eliom (branch shared react),
+Eliom-base-app depends on dbm, dev version of Eliom (branch shared react),
 and dev version of tyxml, js_of_ocaml, ojquery, macaque, ocsigen-widgets and reactiveData.
 Use opam to install it, after pinning the github repositories.
 


yetanotherion:
I wonder whether we souldn't add the dependency in opam instead ?

vasilisp:
Your error message is suspicious. You can't have installed Eliom-Base-App while missing `eliom.server`. Maybe your OPAM environment variables are out of date?

If the server got built without DBM, it must have gotten built with the SQLite Ocsipersist backend instead, which should also work.

yetanotherion:
To add more context:
- I installed eliom-base-app without opam (git clone, make, make install),
- I indeed have the SQLite Ocsipersist backend, but got the error when compiling an application generated with the eba.pgocaml template.

vasilisp:
Can you try with `PERSISTENT_DATA_BACKEND = sqlite` in `Makefile.options`? I don't think dbm is a hard dependency, it should be a recommendation at most. But still, the error message points to other issues.

yetanotherion:
I think you can reproduce the error like that:
$ cd \`ocamlfind query ocsigenserver.ext\`
$ mkdir dbm
$ mv ocsipersist-dbm* dbm
$ cd \<app\>
$ make test.byte
PGHOST=localhost PGDATABASE=test PGPORT=3000 PGUSER=ionalberdi ocsigenserver "-v" -c local/etc/test/test-test.conf
ocsigenserver: ocsigen:main: Fatal - Findlib package ocsigenserver.ext.ocsipersist-dbm not found: maybe you forgot \<findlib path="..."/\>?

dannywillems:
I think it could be closed. The error was related to [this PR in Eliom](https://github.com/ocsigen/eliom/pull/377).

vasilisp:
@dannywillems, not handled by your Eliom PR (because this one is about the OS template), but I think we can close this given that we now use the necessarily-present Postgres backend.

PR #33, opened 2015-10-11
eba_handlers: make forgot_password handler send a mail

yetanotherion:
When clicking in the forgot password button,
an activation link was printed in stdout, but
no mail was sent.

This commit fixes the bug by refactoring the send_act
method.

### DIFF
--- a/src/eba_handlers.eliom
+++ b/src/eba_handlers.eliom
@@ -96,8 +96,8 @@ let send_act msg service email userid =
 let sign_up_handler' () email =
   let send_act email userid =
     let msg =
-      String.concat ", " ["Welcome!\r\nTo confirm your e-mail address";
-                          "please click on this link: "] in
+      "Welcome!\r\nTo confirm your e-mail address, \
+       please click on this link: " in
     send_act msg Eba_services.main_service email userid
   in
   try_lwt
@@ -118,9 +118,8 @@ let sign_up_handler' () email =
 let forgot_password_handler service () email =
   try_lwt
     lwt userid = Eba_user.userid_of_email email in
-    let msg = String.concat "," ["Hi";
-                                 "\r\nTo set a new password";
-                                 " please click on this link: "] in
+    let msg = "Hi,\r\nTo set a new password, \
+               please click on this link: " in
     send_act msg service email userid
   with Eba_db.No_such_resource ->
     Eliom_reference.Volatile.set
--- a/src/eba_handlers.eliom
+++ b/src/eba_handlers.eliom
@@ -82,11 +82,11 @@ let generate_act_key
       with _ -> Lwt.return ());
   act_key
 
-let send_act email userid =
+let send_act msg service email userid =
   let act_key =
     generate_act_key
-      ~service:Eba_services.main_service
-      ~text:"Welcome!\r\nTo confirm your e-mail address, please click on this link: "
+      ~service:service
+      ~text:msg
       email
   in
   Eliom_reference.Volatile.set Eba_msg.activation_key_created true;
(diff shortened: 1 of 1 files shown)


src/eba_handlers.eliom
```diff
@@ -94,6 +94,12 @@ let send_act email userid =
   Lwt.return ()
 
 let sign_up_handler' () email =
+  let send_act email userid =
+    let msg =
+      String.concat ", " ["Welcome!\r\nTo confirm your e-mail address";
+                          "please click on this link: "] in
```
vasilisp: There is no need for `String.concat`. OCaml permits multi-line strings, so this could be just

    "Welcome!\r\nTo confirm your e-mail address \
     please click on this link: "

src/eba_handlers.eliom
```diff
@@ -112,10 +118,10 @@ let sign_up_handler' () email =
 let forgot_password_handler service () email =
   try_lwt
     lwt userid = Eba_user.userid_of_email email in
-    let text = "Hi,\r\nTo set a new password, please click on this link: " in
-    let act_key = generate_act_key ~service ~text email in
-    Eliom_reference.Volatile.set Eba_msg.activation_key_created true;
-    Eba_user.add_activationkey ~act_key userid
```
vasilisp: I haven't run the code. It is nice that you "de-inlined" the version of `send_act` that was hiding inside `forgot_password_handler`, but the real fix is removing the call to `Eba_user.add_activationkey` (which is an orthogonal improvement), right?

src/eba_handlers.eliom
yetanotherion: I just tried it, however it seems that I wouldn't have
the "please"
written in the same column as in "Welcome", 
using multi-line strings

let msg = "Welcome!\r\nTo confirm your e-mail address \
 please click on this link: ";;
val msg : string =
  "Welcome!\r\nTo confirm your e-mail address please click on this link"
# let msg = "Welcome!\r\nTo confirm your e-mail address \ 
                    please click on this link: ";;
val msg : string =
  "Welcome!\r\nTo confirm your e-mail address  \n           please click on this link: "

src/eba_handlers.eliom
yetanotherion: I'll run it again just to be sure, but from the tests I made I had the perception that these modifications fixed the bug.
keep in touch :)

src/eba_handlers.eliom
vasilisp: I suspect that you had a space after `\`; if you remove that, the whitespace on the next line (before "please") will get ignored :).

src/eba_handlers.eliom
yetanotherion: After retrying it, I confirm the fix works in my environment.

src/eba_handlers.eliom
yetanotherion: Indeed, thx :) I pushed a commit to fix that

yetanotherion:
@vasilisp did you have time to review the new commit ?

vasilisp:
I forgot to respond earlier, sorry. The commit looks fine. I will merge later.

vasilisp:
I tried to reproduce the issue before merging, and I can't. The original code works fine for me. I receive a mail (in the presence of a working `/usr/sbin/sendmail`).

Are you sure the issue was not related to your environment? Maybe you fixed your MTA?

What I said about `Eba_user.add_activationkey` above doesn't make sense (I was just confused by a hidden line in GitHub). The new and the old code should be equivalent.

yetanotherion:
I did not have an MTA and was relying in the debug message that was output in stdout, which might be why I did not see it in my previous run. Do you still one this PR ? (as it factorizes the otherwise duplicated
let act_key = generate_act_key ~service ~text email in
Eliom_reference.Volatile.set Eba_msg.activation_key_created true;
Eba_user.add_activationkey ~act_key userid)

vasilisp:
Merged because code duplication is evil :).

PR #34, opened 2015-10-15
Multiplemail

yetanotherion:
Aims at implementing
https://github.com/ocsigen/eliom-base-app/issues/29

There still are missing features:
- add the additional mail edition ui somewhere else (like eba_userbox ?)
- implement rate limiting when sending emails with activation links
- when adding a mail, send an activation link right away without asking the user to click on an "Activate" button
- add some css

Nonetheless, reviews are welcome

### DIFF
--- a/src/eba_db.ml
+++ b/src/eba_db.ml
@@ -233,7 +233,7 @@ module User = struct
                 t.email = $string:email$
          >>
   (* Does not check whether the email is activated *)
-  let update_users_primary_email dbh email =
+  let update_users_primary_email email =
     full_transaction_block (fun dbh ->
        lwt userid = select_user_from_email_q dbh email in
        lwt emails = emails_and_params_of_userid_q userid dbh in
--- a/src/eba_db.ml
+++ b/src/eba_db.ml
@@ -387,6 +387,7 @@ module User = struct
                     t2 in $emails_table$;
                     t1.userid = t2.userid;
                     t2.email = $string:email$;
+                    t2.is_primary = $bool:true$;
             >>
       in
       let (userid, password') = (r#!userid, r#?password) in
--- a/src/eba_db.ml
+++ b/src/eba_db.ml
@@ -196,15 +196,17 @@ module User = struct
         in Lwt.return true
       with No_such_resource -> Lwt.return false)
 
-  let get_users_emails_q userid dbh =
+  let emails_and_params_of_userid_q userid dbh =
     Lwt_Query.view dbh
       <:view< { t.email; t.is_primary; t.is_activated } |
                 t in $emails_table$;
                 t.userid = $int64:userid$;
       >>
 
-  let get_users_emails userid =
-    full_transaction_block (get_users_emails_q userid)
+  let emails_and_params_of_userid userid =
+    lwt res = full_transaction_block (emails_and_params_of_userid_q userid) in
+    let fmt_res x = x#!email, x#!is_primary, x#!is_activated in
+    Lwt.return (List.map fmt_res res)
 
   let activate_email email =
     full_transaction_block (fun dbh ->
@@ -224,7 +226,7 @@ module User = struct
   let update_users_primary_email dbh email =
     full_transaction_block (fun dbh ->
        lwt userid = select_user_from_email_q dbh email in
-       lwt emails = get_users_emails_q userid dbh in
+       lwt emails = emails_and_params_of_userid_q userid dbh in
        let primary = List.filter (fun x -> x#!is_primary) emails in
        lwt () =
            match primary with
--- a/src/eba_db.ml
+++ b/src/eba_db.ml
@@ -108,7 +108,9 @@ let users_table =
 let emails_table =
   <:table< emails (
        email text NOT NULL,
-       userid bigint NOT NULL
+       userid bigint NOT NULL,
+       is_primary boolean NOT NULL,
+       is_activated boolean NOT NULL
           ) >>
 
 let activation_table :
@@ -117,7 +119,7 @@ let activation_table :
     Sql.view =
   <:table< activation (
        activationkey text NOT NULL,
-       userid bigint NOT NULL,
+       email text NOT NULL,
        creationdate timestamptz NOT NULL DEFAULT(current_timestamp ())
            ) >>
 
@@ -228,7 +230,9 @@ module User = struct
         Lwt_Query.query dbh
           <:insert< $emails_table$ :=
                       { email = $string:email$;
-                        userid  = $int64:userid$}
+                        userid  = $int64:userid$;
+                        is_primary = $bool:true$;
+                        is_activated = $bool:true$}
             >>
       in
       lwt () = remove_preregister email in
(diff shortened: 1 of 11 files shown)


src/eba_db.ml
```diff
@@ -117,7 +120,7 @@ let activation_table :
     Sql.view =
   <:table< activation (
        activationkey text NOT NULL,
-       userid bigint NOT NULL,
+       email text NOT NULL,
```
vasilisp: Maybe add an int ID to this table, and use that for referencing its entries from elsewhere?

src/eba_view.eliom
```diff
@@ -190,4 +190,191 @@ let password_form ?a ~service () =
...
+}}
+
+{client{
+
+module ReactList = struct
+    let list t =
+      let open ReactiveData.RList in
+      make_from
+        (React.S.value t)
+        (React.E.map (fun e -> Set e) (React.S.changes t))
```
vasilisp: This functionality is provided by `make_from_s`, and I am not sure it is a good idea to replicate stuff for compatibility with old releases.

src/eba_view.eliom
```diff
@@ -190,4 +190,191 @@ let password_form ?a ~service () =
...
+    let create userido =
+      match userido with
+      | None -> Lwt.return `NotConnected
+      | Some userid ->
+         lwt res = (create_connected userid) in
+         Lwt.return (`Connected res)
+end
+
+let create_input name =
+  Tyxml_js.Html5.(input ~a:[a_input_type `Text; a_placeholder name]) ()
```
vasilisp: Is there a good reason to use `Tyxml_js` instead of the client-side `Eliom_content` modules?

src/eba_db.ml
yetanotherion: The activationkey is for now the primary key of that table. 
I don't see for now, a lot of benefits at putting an id, what do you think ?

src/eba_view.eliom
yetanotherion: I did not know that function :), could you tell me in which eliom (or other) module I could find it ?

src/eba_view.eliom
yetanotherion: Here again, it's due to my lack of knowledge :). I just used Tyxml_js out of the examples I found
of using React
https://github.com/slegrand45/examples_ocsigen/tree/master/jsoo/tic-tac-toe
https://github.com/hammerlab/ketrew/tree/master/src/client-joo
Could you give me suggestions about modules witch which I should replace Tyxml_js ?

src/eba_view.eliom
vasilisp: It is in the `RList` module of `ReactiveData`, like `make_from`. See the [`mli` file](https://github.com/hhugo/reactiveData/blob/master/src/reactiveData.mli) for the API :).

src/eba_view.eliom
vasilisp: Eliom provides its own modules for HTML element generation, including [`Eliom_content.Html5.R`](https://ocsigen.org/eliom/dev/api/client/Eliom_content.Html5.R) that can be used with `React`. The API of the `R` module is very similar to `Tyxml_js.R`, because both are built on top of TyXML.

vasilisp: See ["generating HTML pages"](https://ocsigen.org/eliom/dev/manual/clientserver-html) in the manual, and especially the ["reactive DOM"](https://ocsigen.org/eliom/dev/manual/clientserver-html#reactive) part.

Similar functionality is now available for code inside server sections. There is a [tutorial for that](https://ocsigen.org/tuto/4.2/manual/tutoreact) which describes the basic ideas, but the module names are outdated (`Eliom_csreact` -> `Eliom_shared`, `Eliom_csreact.R` -> `Eliom_content.Html5.R`, plus internal renaming in `Eliom_shared`). It would be nice to use these server-side modules for the new stuff that goes in Eliom-Base-App.

template.distillery/PROJECT_NAME.sql
```diff
@@ -12,12 +12,14 @@ CREATE TABLE users ( -- DEFAULT
 
 CREATE TABLE emails ( -- DEFAULT
        email text primary key, -- DEFAULT
-       userid bigint NOT NULL references users(userid) -- DEFAULT
+       userid bigint NOT NULL references users(userid), -- DEFAULT
+       is_primary boolean NOT NULL,
+       is_activated boolean NOT NULL
 );
```
vouillon: I don't think `email` should remain a primary key. Otherwise, you can register any email address and make it difficult for the actual owner of the address email to use it.

So, we should have the following constraints:
* activated emails should be unique:
````
CREATE UNIQUE INDEX email_unique ON emails (email) WHERE is_activated;
````
* each user should have at most one primary email:
````
CREATE UNIQUE INDEX user_unique ON emails (email) WHERE is_primary;
````

Or maybe just activated email addresses should be unique, so that we can have recovery addresses shared between several accounts?

src/eba_view.eliom
yetanotherion: The first two issues are addressed in https://github.com/yetanotherion/eliom-base-app/commit/351bb7ea7a2433271898ef7c7a19e787a1da2906 and https://github.com/ocsigen/eliom-base-app/commit/2675949c9b9f19da08985b4c24378900489778e0.
I'll have a look at the third one during the next days

template.distillery/PROJECT_NAME.sql
yetanotherion: Good catch !
Regarding the last question:
"Or maybe just activated email addresses should be unique, ..."
I guess you meant
"Or maybe just *primary* email addresses should be unique, ..." didn't you ?

The use case of a parent, setting the same recovery address to all of its children account*, might indeed be worth considering. I'd vote for 
"just *primary* email addresses should be unique".
Do you agree @vouillon ?

*: so that the child can still reset the password and use its account even if he/she forgot his/her password to the application and mail.

vouillon:
Indeed, I meant primary addresses.
So, let's have the following two constraints:
````sql
CREATE UNIQUE INDEX email_unique ON emails (email) WHERE is_primary;
````
````sql
CREATE UNIQUE INDEX user_unique ON emails (userid) WHERE is_primary;
````
(There was a typo on the second constraint above...)

vasilisp:
1. I haven't seen how the feature looks like, but I don't think that putting a big dialog in the userbox would be a good idea. To me it sounds like the user menu (or a page linked to from the user menu) would be a better place.

2. I guess you could just implement a simple and reasonable default limit. Absolute flexibility is beyond the point :).

3. In `Eliom_cscache.find` you explicitly pass a `get_data` function, which by necessity is implemented differently on the server and on the client; the client needs to perform a remote call, and remote calls are generally explicit in Eliom. So the standard way of using `Eliom_cscache` involves this boilerplate, yes. This can definitely be improved. (Sorry if I didn't understand the question.)

balat:
Hello,
I'm currently trying this. Concerning the features, I think it totally responds to the problem. Thank you very much.
To answer your questions:

1. Indeed putting this in the user menu is weird. The most important thing is that there is a function to  create this widget, that programmers can call and put in the page they want. For the default eba app, as the user menu is becoming too big, I suggest we create a "preferences" page, with passwords, emails, etc.
2. This is a complex problem. I think it should probably be addressed more globally. We have the problem with activation emails, but also in many other cases. Actually for any POST request (that will modify the database) ... I think we should first have a look at the solutions used by major Web apps for this. Do you have an idea? 
3. Indeed I hope we can remove this. Can you explain why you need this polling? If I understand well, you want to wait for the parent to be in the page? Why can't you just call the initialization function after generating the box?

Vincent

balat:
Also: I don't understand the optional field "primary email" in lost password form.
I suspect users will be confudes if they see this ....

vasilisp:
Subsumed by #99, closing. Thanks!

PR #36, opened 2015-11-25
add keep me logged in feature in connection form

pwbs:
modified:   src/eba_handlers.eliom
	modified:   src/eba_services.eliom
	modified:   src/eba_session.eliom
	modified:   src/eba_session.eliomi
	modified:   src/eba_view.eliom

### DIFF
--- a/src/eba_handlers.eliom
+++ b/src/eba_handlers.eliom
@@ -131,14 +131,18 @@ let disconnect_handler () () =
   (* SECURITY: no check here because we disconnect the session cookie owner. *)
   Eba_session.disconnect ()
 
-let connect_handler () (login, pwd) =
+let connect_handler () ((login, pwd), keepmeloggedin) =
   (* SECURITY: no check here.
      We disconnect the user in any case, so that he does not believe
      to be connected with the new account if the password is wrong. *)
   lwt () = disconnect_handler () () in
   try_lwt
     lwt userid = Eba_user.verify_password login pwd in
-    Eba_session.connect userid
+    let expire =
+      if keepmeloggedin then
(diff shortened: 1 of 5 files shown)


vasilisp:
The code looks OK. I would find `keep_me_logged_in` (with underscores) more readable; or maybe we can use something sorter like `remember_me` (and "Remember me" for the message).

pwbs:
Strictly speaking, "remember me" has a different meaning: remembering a user does not imply keeping them logged in.

Drup:
@pwbs Well, except that the entire internet call it that way. You are not going to win this one, I'm afraid.

pwbs:
> @pwbs Well, except that the entire internet call it that way. You are not going to win this one, I'm afraid.

Well, I don't care about if the entire internet calls it whatever it calls it, I was just stating a fact. And I don't care about winning or losing either, as there's no fight. To be very clear, I absolutely do not care how that variable is called. It could be `x` or `eat_chocolate`, I couldn't care less.

If you want to change the name of that variable, please feel free to do it, I can only guarantee that I won't care and that I won't be upset about it and that it won't affect anything I do or don't do.

vasilisp:
We all "win" from a cleaner codebase and UI, and nice messages / names contribute to this :). I would go with "Remember me". It is established, it conforms to the unwritten rule of short UI labels, and it appears in the very specific context of a log-in dialog, so the meaning is clear.

Merging to avoid getting in the way of an EBA user, and we can reach consensus on names later.

pwbs:
Thank you @vasilisp.

@vasilisp @Drup Please feel free to change that "keep me logged in" stuff in any way that suits you.

PR #40, opened 2016-01-12
add an explicit project tmp dir: ELIOMTMPDIR

pwbs:
modified:   template.distillery/Makefile
modified:   template.distillery/Makefile.options


vasilisp:
I am not sure I understand. Does any part of Eliom/EBA know about and rely on `$ELIOMTMPDIR`? If not, why should it be created by the build system?

pwbs:
If I understood the problem correctly (perhaps let @vouillon and/or @balat confirm this...) 
it's because we should have a placeholder for a temporary variable so that the user is not tempted by using TMPDIR instead, because TMPDIR is a special variable and as such it's exported in the environment of processes created by the Makefile.

pwbs:
The choice of `:= tmp` might not be perfect... some people might prefer `:= var/tmp`, I don't know.

http://unix.stackexchange.com/questions/30489/what-is-the-difference-between-tmp-and-var-tmp

PR #47, opened 2016-04-04
Used last version of ocsigen-toolkit (in replacement to some ocsigen-widgets module)

sagotch:
This PR also remove default dependency to fontawesome. Is it okay for you @balat @vasilisp?
Side effect of switching to `ot_picture_uploader` is the drop of `jcrop` dependency.

I think font-awesome is not necessarily the way to go anymore (I mean: not for everybody). A lot of free icon sets are available now, and if one want to use one of them it is very easy to add css and font files.

If you want to, I can amend this to restore the dependency to font awesome and restore the `PROJECT_NAME_icons.eliom` file. My feeling was just that it does not have to be shipped with eba.


vasilisp:
I wasn't persuaded that we need Font Awesome anyway, so fine with me. I am also happy that you reduced the dependence on jQuery. I will take a look at the code later.

balat:
We don't need fontawesome but we need default icons.
Where do they come from now?

sagotch:
Ot_icon have some pure css icons. I can improve the set of icons provided if needed.

sagotch:
I just rebased it upon master.

If someone has time to review this it will avoid me to maintain the branch for too long :)

vasilisp:
I have looked at it, and I have nothing interesting to say. As long as you are ready to use this branch for BeSport, we can merge it.

sagotch:
I will rebase -i to re-add fontAwesome since @balat prefers it to be included by default.

sagotch:
It's now ok for me.

template.distillery/PROJECT_NAME_page.eliom
```diff
@@ -41,6 +39,10 @@ let%shared the_local_css = [
   ["font-awesome.css"];
   ["jquery.Jcrop.css"];
   ["eba.css"];
+  ["ot_icons.css"] ;
+  ["ot_picture_uploader.css"];
+  ["ot_popup.css"];
+  ["ot_spinner.css"]
```
vasilisp: We need a semicolon here.

vasilisp:
Can you do one last rebase to fix the conflicts? Then I will merge it. Thanks!

sagotch:
Rebased, seem ok to me except a css issue (see https://github.com/ocsigen/eliom-base-app/issues/56). I only really tested the profile picture upload, which is working.

PR #52, opened 2016-04-13
Makefile dependency to ocsigen-toolkit not ocsigen-widgets

jrochel:


vasilisp:
As far as I can tell, the dependency on `ocsigen-widgets` is not entirely gone. `grep Ow_ src/*` returns lots of stuff. But in any case, I think we won't need an explicit dependency to either package, given that the right one will be dragged in as dependency of `eliom-base-app.{client,server}`.

PR #70, opened 2016-06-20
add class 'eba-client-app' to html in eliom.html

pwbs:
modified:   template.distillery/mobile!eliom.html.in


pwbs:
(because it's possible that the html node isn't generated along with the page, as the new page might use the already existing one)

PR #80, opened 2016-07-21
Eba notif

canonici:
Updated to use eliom_notif.


vasilisp:
@balat, do you want to test this before we merge it? It looks OK to me.

@canonici, I feel tempted to squash the commits because the history is not all that clean (all the fix and corr are for what?). Do you want to force-push a cleaned-up history? Or I can do it and merge manually, as you wish.

pwbs:
> @canonici, I feel tempted to squash the commits because the history is not all that clean (all the fix and corr are for what?). Do you want to force-push a cleaned-up history? Or I can do it and merge manually, as you wish.

I concur!

canonici:
I squashed the commits.

vasilisp:
We need to rebase this on top of master, and re-implement `unlisten_user` (#104) in Eliom. Not entirely trivial because of the change away from volatile references (whose consequences I don't exactly understand).

vasilisp:
Superseded by #314.

PR #86, opened 2016-07-22
DO NOT MERGE Trying to restart the app after connection

balat:
I'm trying to make the client app (mobile or not) restart when a user connects/disconnect.

Here is my current version that needs to be discussed.

 - I'd like to use Eliom_registration.Unit.register but it seems not implemented client side?
 - Is it possible to use Eliom_registration.Any for both sides in that case?
 - This version does not work well because there is also a restart triggered by Eba_comet.handle_message
May be it's enough to keep this (but the restart should probably be reimplemented for the mobile app, as we want to restart from eliom.html, not the current page). But anyway, i'd like to understand how to do that without this Eba_comet restart.

Also TODO: make possible to restart the app at any page (eliom.html + change_page_uri to the right URL ? ...) 

@vasilisp @vouillon any ideas?

### DIFF
--- a/src/eba_handlers.eliom
+++ b/src/eba_handlers.eliom
@@ -161,27 +161,61 @@ let%client disconnect_handler_rpc' =
        disconnect_handler_rpc')
 let%client disconnect_handler () v = disconnect_handler_rpc' v
 
+let%shared restart () =
+  (* We ask the client to restart.
+     On a Web app, it is just reloading the page.
+     On a mobile app, we want to restart from eliom.html.
+  *)
+  ignore [%client
+    (print_endline "restarting";
+     if Eliom_client.is_client_app ()
+     then Eliom_client.exit_to ~absolute:false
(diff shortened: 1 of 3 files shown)


vasilisp:
> I'd like to use Eliom_registration.Unit.register but it seems not implemented client side?

Not implemented yet, just because I haven't seen them used (and there is a [discouraging server-side comment](https://github.com/ocsigen/eliom/blob/master/src/lib/eliom_registration.server.ml#L508)). The idea is that we have an "action" that performs a side-effect, and we do nothing afterwards? That's a 2-minute addition if you want it :).

vasilisp:
Leaving the conflict with comet aside, I don't see anything wrong with doing a `change_page` towards `eliom.html`. If you want to go directly to a specific page, you can use a global JS variable to store the target URL, and jump to it at startup (via `change_page_uri`). You have privately sent me a code sample that does a similar thing in `handle_initial_url` :).

PR #91, opened 2016-08-04
Eba_lib: base_and_path_of_url: useful when you want to split

dannywillems:


dannywillems:
@vasilisp do you think is useless or you forgot this PR? :)

pwbs:
@dannywillems meanwhile, your branch has conflicts... ;)

dannywillems:
@pwbs I think it will be better to close it and I will add it when I will PR OAuth2 and OpenID Connect.

vasilisp:
@dannywillems, the code looks good, but I am not exactly sure what `Eba_lib` should contain. Also, I would be a bit reluctant to add new URL manipulation functions in a new place; we already have multiple `Url` modules in Ocsigenserver and Eliom and inconsistencies between client and server.

PR #95, opened 2016-08-12
Eba_tips: used a function to generate tip content in order to avoid actually building it if not necessary.

sagotch:


vasilisp:
Looks good. Have you updated the BeSport code to use this? We can merge it whenever you are ready.

sagotch:
I do have a branch synchronized with this :+1:

src/eba_tips.eliom
```diff
@@ -117,19 +117,20 @@ let%shared block ?(a = []) ~name ~content () =
...
     then Lwt.return None
     else begin
-      let close_button = Ot_icons.D.close () in
+      let box_ref = ref None in
+      let close = [%client (fun () ->
+        let () = match !(~%box_ref) with
+          | Some x -> Manip.removeSelf x
+          | None -> () in
+        Lwt.async (fun () -> set_tip_seen ~%name);
+        Lwt.return () : _ -> _) ] in
```
vasilisp: Upon further inspection: why does `close` need to return an Lwt value?

src/eba_tips.eliomi
```diff
@@ -51,7 +51,8 @@ val bubble :
   ?width:int ->
   ?parent_node:[< `Body | Html_types.body_content ] Eliom_content.Html.elt ->
   name:string ->
-  content: Html_types.div_content Eliom_content.Html.elt list ->
+  content:((unit -> unit Lwt.t) Eliom_client_value.t
```
vasilisp: Can you document the `(unit -> unit Lwt.t) Eliom_client_value.t` argument?

src/eba_tips.eliom
sagotch: Very good question, I have no idea why the Lwt thing is here. Probably needed in a first version, I'll remove it!

src/eba_tips.eliomi
sagotch: Sure

src/eba_tips.eliom
sagotch: Or maybe the question is: why does `close` is wrapped in a `Lwt.async` call? Shouldn't we wait for the action to complete?

i.e. replace 
```
Lwt.async (fun () -> set_tip_seen ~%name) ; Lwt.return () 
```
with
```
set_tip_seen ~%name
```

sagotch: I pushed a commit unwrapping the closing function call, is it okay to do that?

vasilisp: Thanks, looks good now: `close` truly does do something asynchronous, as opposed to just scheduling another asynchronous task.

PR #81, opened 2016-07-21
Eba template

canonici:
Update of the eba template to make it responsive and a little bit more _"stylish"_ .

PR #97, opened 2016-08-19
rename ocsigen-start

canonici:
renamed every occurrence of eba to os


vasilisp:
This will break the template as far as I can see, but that's OK, we will get back a working template with a rebased #81 .

pwbs:
Don't forget to make http://www.ocsigen.org/ocsigen-start work... ;)

vasilisp:
@balat @vouillon can you start your migration to the `Os_` modules so that we can proceed with the renaming? Thanks!

pwbs:
I think you should rename before we switch... Otherwise, to switch, we'd have to pin the branch of the PR (or a clone of it).

vasilisp:
OK, then we need one of @balat @vouillon @Drup to rename the repo. I don't have permissions to do that. We can merge the PR immediately thereafter, and then you switch.

pwbs:
I think it should be more like 
- make a new repo, name it ocsigen-start, push the stuff there
- display a big redirection link on this repo, then switch it to read-only

Drup:
Github automatically keep a redirection for renamed projects. No need to do a complicated scheme, especially for a project that was never properly released.

Drup:
Done.

pwbs:
Oh, nice, I didn't know! :)

(However I suppose it doesn't happen when it's moved from an account to another?)

Drup:
If you move from one account to another, the old one is mark as a forked, iirc.

pwbs:
Please merge! ;-)

PR #98, opened 2016-08-22
Redirect to home page after activation

vasilisp:
This PR restores the server-side behavior of `Eba_handlers.activation_handler` before #88. I think it doesn't interfere with the intended client-side effect of #88. 

The server-side `Eba_handlers.activation_handler` needs to redirect to the main page. If it doesn't, we stay on a URL with the activation key in it; once the user fills-in their data, the action takes us to the same URL that has a now-invalid activation key.

CC @vouillon @canonici @balat

### DIFF
--- a/src/os_handlers.eliom
+++ b/src/os_handlers.eliom
@@ -194,7 +194,7 @@ let%client connect_handler_rpc =
        connect_handler_rpc')
 let%client connect_handler () v = connect_handler_rpc v
 
-let activation_handler_common ~restart ~akey =
+let activation_handler_common ~akey =
   (* SECURITY: we disconnect the user before doing anything. *)
   (* If the user is already connected,
      we're going to disconnect him even if the activation key outdated. *)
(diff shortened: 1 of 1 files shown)


vasilisp:
I can rebase this after we merge #97, probably easier in that order.

pwbs:
> The server-side Eba_handlers.activation_handler needs to redirect to the main page. If it doesn't, we stay on a URL with the activation key in it; once the user fills-in their data, the action takes us to the same URL that has a now-invalid activation key.

I think that redirecting to the main page is wrong. But of course, keeping an invalid key in the URL is even worse... But it's because of the way the keys currently work.

One fundamental problem is that activations keys are used for different things that shouldn't be that mixed up:
- account activation
- password reset
- and other things too (when we send an email to someone and we log that person in regardless of whatever is happening).

On BS, when we invite someone to an event, we currently provide an activation link. It's wrong to do it this way, so we're (I am, actually) changing that.

Also, it doesn't make sense that when someone arrives with an invalid key, they get logged out (not to say "kicked out")... Anyways, I'm currently working on fundamentally changing a few things regarding this matter...

src/eba_handlers.eliom
```diff
@@ -203,22 +203,29 @@ let activation_handler_common ~restart ~akey =
...
       Eba_userbox.activation_key_outdated true;
     Eba_msg.msg ~level:`Err ~onload:true
       "Invalid activation key, ask for a new one.";
-    Lwt.return ()
+    Lwt.return `NoReload
 
 let%server activation_handler akey () =
-  let%lwt () = activation_handler_common ~restart:false ~akey in
-  Eliom_registration.Action.send ()
+  let%lwt a = activation_handler_common ~restart:false ~akey in
```
pwbs: Note that `~restart` is never used by `activation_handler_common`... In a branch of mine, I've removed `~restart` from `activation_handler_common`.

vasilisp:
@pwbs, I haven't thought about how activation keys should work, and I am glad you do. The scope of this PR is limited to a small regression in #88. We can merge it if it helps or don't bother with it and wait for your reworked version. As you (BeSport) wish :).

pwbs:
Ideally, I think this should be merged... It'll just be a bit harder to merge but this shouldn't be a reason to delay fixing a regression.

I'm hoping to finish what I'm doing with those activation links this week but I keep meeting new tricky points about it so I'm not so sure about when I get there! :-/

src/os_handlers.eliom
```diff
@@ -203,22 +203,29 @@ let activation_handler_common ~restart ~akey =
...
     let%lwt () = Os_session.connect userid in
-    Lwt.return ()
+    Lwt.return `Reload
   with Os_db.No_such_resource ->
     Eliom_reference.Volatile.set
       Os_userbox.activation_key_outdated true;
     Os_msg.msg ~level:`Err ~onload:true
       "Invalid activation key, ask for a new one.";
-    Lwt.return ()
+    Lwt.return `NoReload
```
pwbs: @vasilisp I'm wondering why it's `` `NoReload`` here instead of `` `Reload``. Using `` `NoReload`` mostly means that the user will keep that invalid key in the URL, which is not very useful. Did I miss something?

pwbs: Ok, I missed something!

PR #99, opened 2016-08-23
multimail

canonici:
Support for multiple emails for one user + little rewrite of os_db.ml

PS:  .sql files need to have a [email citext NOT NULL,] field for the activation table for this to work.

### DIFF
--- a/src/os_db.ml
+++ b/src/os_db.ml
@@ -21,6 +21,7 @@
  *)
 
 exception No_such_resource
+exception Main_email_removal_attempt
 
 let (>>=) = Lwt.bind
 
@@ -113,6 +114,7 @@ let users_table =
        userid bigint NOT NULL DEFAULT(nextval $users_userid_seq$),
        firstname text NOT NULL,
        lastname text NOT NULL,
+       main_email citext NOT NULL,
        password text,
        avatar text
           ) >>
@@ -131,6 +133,7 @@ let activation_table :
   <:table< activation (
        activationkey text NOT NULL,
        userid bigint NOT NULL,
+       email citext NOT NULL,
        creationdate timestamptz NOT NULL DEFAULT(current_timestamp ())
            ) >>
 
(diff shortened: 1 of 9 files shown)


src/os_user.eliomi
```diff
@@ -33,14 +33,14 @@ val is_complete : t -> bool
...
+val add_activationkey : act_key:string -> int64 -> string -> unit Lwt.t
 val verify_password : email:string -> password:string -> int64 Lwt.t
 
 (** returns user information.
     Results are cached in memory during page generation. *)
 val user_of_userid : int64 -> t Lwt.t
 
-val userid_of_activationkey : string -> int64 Lwt.t
+val userdata_of_activationkey : string -> (int64 * string) Lwt.t
 (** Retrieve an userid from an activation key. May raise [No_such_resource] if
```
pwbs: please update the comments... ;-)

src/os_user.eliomi
```diff
@@ -33,14 +33,14 @@ val is_complete : t -> bool
 
 val emails_of_user : t -> string Lwt.t
 
-val add_activationkey : act_key:string -> int64 -> unit Lwt.t
+val add_activationkey : act_key:string -> int64 -> string -> unit Lwt.t
```
pwbs: perhaps labelling would be better than having int64 and string

```ocaml
val add_activationkey : act_key:string -> userid:int64 -> email:string -> unit Lwt.t
```

src/os_handlers.eliomi
```diff
@@ -10,6 +10,8 @@ val sign_up_handler : unit -> string -> unit Lwt.t
 val activation_handler :
   string -> unit -> Eliom_registration.Action.result Lwt.t
 
+val add_mail_handler : int64 -> unit -> string -> unit Lwt.t
```
pwbs: is the parameter of type `unit` really necessary?

src/os_user.eliomi
pwbs: Actually, the name of that function is a bit weird, because it's not so clear that it's returning the userid plus the email addr. of the user. Perhaps name it userid_and_email_of_activationkey or else return more complete userdata in a record?

canonici: I updated the comments and changed the name of the function. "userdata" was a bit misleading.

src/os_handlers.eliomi
canonici: No it isn't! Gone.

pwbs: But shouldn't it be
```ocaml
val add_mail_handler : (int64 * string) -> unit -> unit Lwt.t
```
then?  (I don't know)

canonici: I changed it a little bit, I was checking twice if the user was connected.

`val add_mail_handler : unit -> string -> unit Lwt.t`

It looks better.

pwbs:
Please don't delete anything from the activation table...

Perhaps we should work on a common branch so that you can be aware of what I'm doing?

pwbs:
@canonici  

Voici une suggestion:

faire une nouvelle branche à partir de la mienne
```
git checkout b6ab8c60ab97016414d19e29454f2885e660eead
git checkout -b multiemail-and-activation
git pull git@github.com:pwbs/ocsigen-start.git os-activ --rebase
```
puis refaire les patchs que t'as fait aujourd'hui à partir de cette branche...

Si t'as d'autres idées, je suis à l'écoute.

vasilisp:
> Perhaps we should work on a common branch so that you can be aware of what I'm doing?

I think you should :). Maybe close one of the two PRs and collaborate on a single PR? What the two of you are doing seems strongly related, e.g., @canonici was explaining to me that a new e-mail requires re-activating a previously-activated account, and this takes us to @pwbs territory.

canonici:
I'm merging our changes in a new branch.

canonici:
Ok I merged your changes. Now I can keep track of what you are doing.

pwbs:
> I think you should :). Maybe close one of the two PRs and collaborate on a single PR? What the two of you are doing seems strongly related, e.g., @canonici was explaining to me that a new e-mail requires re-activating a previously-activated account, and this takes us to @pwbs territory.

So, the principle of my PR is to keep all activation keys and they are no longer actual plain "activation" keys. Some will log visitors in, some won't. Each key should be given a specific action. It should be possible to add new actions for any ocsigen-start user. An overused activation key shouldn't always be just an invalid key, instead sometimes different actions will be taken regarding the validity of the key. Each key may be used a number of times (by default, it's 1). People will choose what to do when the number reaches 0. (The behaviour when it's less than 0 is undefined...) To make a key usable an unlimited amount of times, just give it `Int64.max_int`. It's not ideal but it's simple to implement and maintain, and it should work. (Using a smaller field, like an int32 or int16 or int8 would bring other problems, using more than 1 field would add complexity.)

> Ok I merged your changes. Now I can keep track of what you are doing.

Thanks :)

pwbs:
Please may someone review and merge this? :)

vasilisp:
@pwbs I will review the code today, but it is fairly invasive and needs testing on your side. You have done some testing I guess? Am I free to push the green button after my review?

pwbs:
Well, the "activation key" part seems to work like intended.

I have not explicitly tested the multi mail part (I've installed it, I'm using an ocsigen-start that contains it, but I've not done any particular test regarding that).

src/os_db.ml
```diff
@@ -158,294 +165,330 @@ let preregister_table =
...
+
+  let one f ~success ~fail q =
+    f q >>= function
+    | r::_ -> success r
+    | _ -> fail
+
+  let all f ~success ~fail q =
+    f q >>= function
+    | [] -> fail
+    | r -> success r
```
vasilisp: To be honest, I do not like `one` and `all`. We don't save much code (given that we have to lambda-abstract the bodies of `success`, `fail`, and `f q`) and it is not immediately visible what success and fail mean (it is more like found and not-found).

src/os_db.ml
```diff
@@ -158,294 +165,330 @@ let preregister_table =
...
-                        password = of_option $password$
-                       } |
-                       d.userid = $int64:userid$
-             >>
-      ))
+    if password = Some "" then Lwt.fail_with "empty password"
+    else
+      let password = password => (
+	(fun p _ -> as_sql_string @@ fst !pwd_crypt_ref p),
+	(fun _ -> password_of)
```
vasilisp: Using `=>` requires as much code as simply pattern matching on password, so why not just use the built-in and known to all construct?

vasilisp:
Looks OK overall. We can accept it as-is and progressively move away from the `one` and `all` idiom.

@canonici, a history cleanup would be greatly appreciated (maybe not full, but at least squashing the fixes into the parent commits that they fix).

canonici:
Ok I did a history cleanup, hopefully I didn't forget to add something important.

vasilisp:
Merci @canonici !

PR #104, opened 2016-08-29
Adding Os_notif.unlisten_user

balat:


src/os_notif.eliom
```diff
@@ -146,6 +146,21 @@ VVV See if it is still needed
...
+  let unlisten_user ~userid (id : A.key) =
+    let state =
+      Eliom_state.Ext.volatile_data_group_state
+        ~scope:Eliom_common.default_group_scope
+        (Int64.to_string userid)
+    in
+    Eliom_state.Ext.iter_volatile_sub_states
+      ~state
+      (fun state ->
+        Eliom_state.Ext.iter_volatile_sub_states
```
vasilisp: Are the nested `iter_volatile_sub_states` on purpose? And if so, could you provide a comment explaining why?

balat: This is a way to iter on all processes of all sessions of a group.
There is an explanation in the mini-howto (but hidden at the very end of tutorials):
http://ocsigen.org/tuto/5.0/manual/how-to-iterate-on-all-sessions-for-one-user-or-all-tabs
I added some comments in the code.

vasilisp:
@balat, merge it if you need it immediately, but this code will be replaced really soon (#80). We need to add an `unlisten_user` in `Eliom_notif` and OS will "inherit" it.

balat:
Ok thanks. I merge it.
Try not to remove the function ;)

PR #106, opened 2016-08-31
fix behaviour when multiple use of plain activation key

pwbs:
- before: reusing an plain activation key leads to an error
- after: reusing an plain activation key just ignores the key
- justification: when an plain activation key has been used before, it
  means that the account was activated. It's okay to just ignore the
  key if it's used again, since the associated action cannot be done
  again anyway, nor can it be undone.

- By "plain activation key", I mean a key which purpose is to activate
  an account. It is the default kind of key. Other kinds of keys
  include, for instance, password reset keys.

	modified:   src/os_db.ml
	modified:   src/os_handlers.eliom

### DIFF
--- a/src/os_handlers.eliom
+++ b/src/os_handlers.eliom
@@ -207,8 +207,12 @@ let activation_handler_common ~akey =
      TODO: do not disconnect users if we relog them with the same userid.
   *)
   try%lwt
-    let%lwt {userid; email; validity; action = _; data = _; autoconnect} =
+    let%lwt {userid; email; validity; action; data = _; autoconnect} =
       Os_user.get_activationkey_info akey in
+    let%lwt () =
+      if action = "activation" && validity <= 0L then
+        Lwt.fail Os_db.Account_already_activated
+      else Lwt.return_unit in
     let%lwt () =
       if validity <= 0L then
         Lwt.fail Os_db.No_such_resource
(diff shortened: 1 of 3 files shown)


src/os_handlers.eliom
```diff
@@ -207,9 +207,13 @@ let activation_handler_common ~akey =
      TODO: do not disconnect users if we relog them with the same userid.
   *)
   try%lwt
-    let%lwt {userid; email; validity; action = _; data = _; autoconnect} =
+    let%lwt {userid; email; validity; action; data = _; autoconnect} =
       Os_user.get_activationkey_info akey in
     let%lwt () =
+      if action = "activation" && validity <= 0L then
```
vasilisp: Not critical for this PR, but could you move to a variant (possibly polymorphic) for the actions? One class of possible bugs less.

pwbs: I agree and thought about that before, and then gave up... It seemed to be not trivial, because I don't want OS to list all possible actions: I want the users of OS to be able to define and implement their own actions.

I'll think again about how to do this.

vasilisp: I guess you can use a `` `Custom of string`` constructor for the user-provided actions. `` `Custom s`` can be stored as `s` in the DB, and when reading something from the DB, we will be using `` `Custom`` as a fallback case for the unknown actions.

pwbs: Done...

PR #108, opened 2016-08-31
Update installation instructions

dannywillems:
It's more convenient to have to copy-paste the command instead of pinning manually each repository.

I just tried on a fresh installation on Mac OS X.

I don't give the instruction to install ImageMagick because it's done on the website.

I removed the old installation instructions because it was out-dated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,20 @@ modules from Ocsigen-start.
 
 ###<a id="install"></a>Installation
 
-Ocsigen-start depends on Eliom >= 5.0, ojquery (development version),
-Macaque, ocsigen-widgets and reactiveData. You can use OPAM to install
-everything; you need to pin the repositories for ocsigen-widgets,
-ojquery, and Ocsigen-start itself.
+Ocsigen-start has a list of dependencies. All can be installed using opam. Here the commands:
+```
+opam pin add ocsigenserver https://github.com/ocsigen/ocsigenserver.git
+opam pin add js_of_ocaml https://github.com/ocsigen/js_of_ocaml.git
+opam pin add deriving https://github.com/ocsigen/deriving.git
+opam pin add reactiveData https://github.com/ocsigen/reactiveData.git
+opam pin add eliom https://github.com/ocsigen/eliom.git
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,10 +33,20 @@ modules from Ocsigen-start.
...
+opam pin add deriving https://github.com/ocsigen/deriving.git
+opam pin add reactiveData https://github.com/ocsigen/reactiveData.git
+opam pin add eliom https://github.com/ocsigen/eliom.git
+opam pin add ocsigen-toolkit https://github.com/ocsigen/ocsigen-toolkit.git
+opam pin add ojquery https://github.com/ocsigen/ojquery.git
+opam pin add ocsigen-widgets https://github.com/ocsigen/ocsigen-widgets.git
+opam pin add ocsigen-start https://github.com/ocsigen/ocsigen-start.git
+```
+
+Ocsigen-widgets depends on the [OCaml binding to ImageMagick C library](https://github.com/besport/ocaml-imagemagick) so you need to install the [C library](http://www.imagemagick.org/script/index.php).
```
vasilisp: Maybe add `--no-action` (and optionally `-y`) to all but the last one? If the dependency order changes (or if the user doesn't follow your order) the process may involve recompilation of the same things.

vasilisp: For `imagemagick` (which BTW creates both installation trouble and security threats, so we should consider eliminating it) the Debian/Ubuntu dependencies are `libgraphicsmagick1-dev` and `libmagickcore-dev`. Much easier and cleaner than going through the website.

dannywillems: I tried with --no-action but the compilation fails while installing text because there were a dependency with camlp4 but camlp4 wasn't yet installed. I needed to relaunch `opam install ocsigen-start`.

dannywillems: For imagemagick, the dependencies are in ow_pic_uploader.eliom to resize, crop, read, write an image and get some info of the image.
Is there an alternative?

vasilisp:
Merging. For the `text` issue, I guess we could fix the dependencies in `opam-repository`.

PR #109, opened 2016-08-31
Avoid to stop building if platform is already added

dannywillems:


dannywillems:
Something wrong with this PR?

pwbs:
@dannywillems I think the patch isn't quite right. For instance, if you don't have the rights to write in $(CORDOVAPATH), with your patch, it won't fail. If you want it not to fail when the platform already exists, I think you have to do something a little more complex...

pwbs:
perhaps merge it and make it better later...

dannywillems:
@pwbs I think a FIXME/NOTE message in comment before could be a compromise. What do you think?

dannywillems:
I can't reproduce the error so I close the error and I will reopen it when I will be able to reproduce it.

PR #110, opened 2016-09-01
sed: BSD compatibility

dannywillems:
sed on Mac OS needs the extension with the option -i.
I tried on a Debian and it's also working.

### DIFF
--- a/template.distillery/Makefile.style
+++ b/template.distillery/Makefile.style
@@ -27,7 +27,7 @@ $(CSSDIR):
 $(DEST): $(LOCAL_CSS) | $(CSSDIR)
 	HASH=`cat $< | md5sum | cut -d ' ' -f 1` && \
 	cp $< $(CSS_PREFIX)_$$HASH.css && \
-	sed -i '1s/^/@charset "UTF-8";/' $(CSS_PREFIX)_$$HASH.css && \
+	sed -i .css '1s/^/@charset "UTF-8";/' $(CSS_PREFIX)_$$HASH.css && \
 	ln -sf $(PROJECT_NAME)_$$HASH.css $@
 # Charset is necessary for iOS.
 # Including it in scss does not work because sass removes it.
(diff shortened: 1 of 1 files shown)


vasilisp:
I don't understand how this could be working on Debian, are you sure? `.css` is parsed as an argument detached from `-i`, so it becomes the `sed` expression, and it happens to be invalid.

```
sed: -e expression #1, char 1: unknown command: `.'
```

template.distillery/Makefile.style
```diff
@@ -27,7 +27,7 @@ $(CSSDIR):
 $(DEST): $(LOCAL_CSS) | $(CSSDIR)
 	HASH=`cat $< | md5sum | cut -d ' ' -f 1` && \
 	cp $< $(CSS_PREFIX)_$$HASH.css && \
-	sed -i '1s/^/@charset "UTF-8";/' $(CSS_PREFIX)_$$HASH.css && \
+	sed -i .css '1s/^/@charset "UTF-8";/' $(CSS_PREFIX)_$$HASH.css && \
```
pwbs: simplest way is probably to write `sed -i -e '1s/^/@charset "UTF-8";/' $(CSS_PREFIX)_$$HASH.css && \ `

pwbs:
@dannywillems It's more BSD compatibility than mac OS compatibility... ;)

template.distillery/Makefile.style
dannywillems: @vasilisp my bad, I was on the wrong session when I ~~tried~~ thought I tried on Debian --'.

dannywillems:
@vasilisp @pwbs -e works.

pwbs:
normally 
`echo pppppPPP > d ;  sed -i.old -e 's/p/P/g' d`
gives the same results with BSD sed and GNU sed. So, perhaps use this?

vasilisp:
@dannywillems , don't you end up with an auxiliary file with the `-e` suffix? See discussion in https://github.com/ocsigen/eliom/commit/54240d4ca2f617206b8158615ec42dcf104e3667 .

pwbs:
otherwise, don't use -i (but use a temporary file) or detect version of sed as in
```bash
if (( sed --version|grep -q GNU )); then
SEDI='sed -i '
else
SEDI='sed -i ""'
fi
# and then use $SEDI ...
```

dannywillems:
@pwbs seems to have no --version on BSD. So need to detect the error.

```
sed --version
sed: illegal option -- -
```

pwbs:
just send error to void: `2> /dev/null`

PR #113, opened 2016-09-05
Server-side interface to GCM to send push notifications.

dannywillems:
Add the module Os_push_notifications which allows to send push notifications (with the binding to Cordova plugin phonegap-plugin-push https://github.com/dannywillems/ocaml-cordova-plugin-push) using Google Cloud Messaging (GCM). GCM allows to send notifications to Android and iOS devices (not tested with iOS but must work if we trust the official documentation).

The module is named Os_push_notifications (and not os_gcm_notifications for example) because it aims to abstract the push notifications service we use. Furthermore, GCM is migrating to FCM (Firebase Cloud Messaging) so it will be easier to migrate to FCM without changing the interface.
I didn't try FCM but it is in progress.

The payloads come from [this link](https://github.com/phonegap/phonegap-plugin-push/blob/master/docs/PAYLOAD.md).
Picture, message, title, sound, logo and priority has been tested. Actions seems not working but I need to investigate deeper. Other payload will be tested soon.

~~Another issue seems to be the notification stacking: all notifications are stacked, even if we set a different collapse key. Seems to come from the *data* payload. **Work in progress.**~~

### DIFF
--- a/Makefile.options
+++ b/Makefile.options
@@ -38,7 +38,7 @@ TEMPLATE_DIR := template.distillery
 TEMPLATE_NAME := os.pgocaml
 
 # OCamlfind packages for the server
-SERVER_PACKAGES := lwt.ppx js_of_ocaml.deriving.ppx text calendar safepass  ocsigen-widgets.server ocsigen-toolkit.server magick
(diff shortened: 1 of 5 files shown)


dannywillems:
Stack issue resolved.

PR #116, opened 2016-09-07
Extract some values from the config.xml.in to Makefile.mobile

dannywillems:
In this way, the developer doesn't need to modify the config.xml.in file but only Makefile.PROJECT_NAME.
However, I think it would be better to put all variables the user can changed in a separate Makefile (like in the Makefile.local). If you agree with that, I need to change this PR by checking if the variables are already defined.


dannywillems:
if #123 is merged, this PR must be changed.

PR #123, opened 2016-09-09
Split Makefile.PROJECT_NAME in Makefile.db and Makefile.options.

dannywillems:
This PR does several things.
1. `Makefile.PROJECT_NAME` contained the database rules and the database settings/variables. @balat wanted to have a separated Makefile.db with the database rules. So `Makefile.db` contains all database rules initially defined in `Makefile.PROJECT_NAME`.
2. Move all variables in `Makefile.options` (from database and style makefiles, no more in other).
3. As the initial `Makefile.PROJECT_NAME` contains only rules for database and variables for mobile ID, it is now empty. @balat suggests to keep it and it will be in this makefile the developer will define his own rules.
4. I began to comment all makefiles. For the moment, I described what is the purpose of `Makefile.PROJECT_NAME` (as described above), the main makefile, `Makefile.db` and database variables in `Makefile.options`.
5. I still found an occurrence of *EBA* and *ELIOM-BASE-APP* and the previous PR was already merged.
6. And, last but not least, added `pg_ctlcluster` for Debian :smiley: , which is a wrapper to `pg_ctl` on Debian (it is not recommended to use `pg_ctl`directly as said on [this StackOverflow link](http://dba.stackexchange.com/questions/72580/missing-the-pg-ctl-package-in-postgres-9-3-installation). For the moment, we search pg_ctl and after pg_ctlcluster. As it's not recommended, do we invert?

If this pull request is merged, I need to change #116.


dannywillems:
@balat : is it normal the rebase causes so many commits?

PR #130, opened 2016-09-13
From bs Makefile.bs: add db-snapshot rule.

dannywillems:
- Use DB_SNAPSHOT instead of DBSNAPSHOT and put this variable in Makefile.options.
- Use a variable for pg_dump.

### Makefile.db reorganisation
- Move pg_dump and pg_ctl at the top (because all variables must be on the top).
- Remove an useless empty new line.

### DIFF
--- a/template.distillery/Makefile.db
+++ b/template.distillery/Makefile.db
@@ -88,11 +88,6 @@ pg_ctl  = $(shell \
 $(PSQL_DIR):
 	-mkdir -p $@
 
-pg_ctl = $(shell \
-           sh -c "which pg_ctl" || \
-           sh -c "which pg_ctlcluster" || \
-           sh -c "if [ -f pg_ctl ]; then (echo ./pg_ctl) else (echo '') fi")
-
 ifeq ($(pg_ctl),)
 $(error "no 'pg_ctl' binary found")
 endif
--- a/template.distillery/Makefile.db
+++ b/template.distillery/Makefile.db
@@ -30,7 +30,19 @@ ELIOMDEP          := $(ENV_PSQL) $(ELIOMDEP)
 # should be the same as ENV_PSQL)
 OCSIGENSERVER     := $(OCSIGENSERVER)
 OCSIGENSERVER.OPT := $(OCSIGENSERVER.OPT)
+
 ## ---------------------------------------------------------------------
+
+##----------------------------------------------------------------------
+
+pg_dump = pg_dump
+
+pg_ctl  = $(shell \
+           sh -c "which pg_ctl" || \
+           sh -c "which pg_ctlcluster" || \
+           sh -c "if [ -f pg_ctl ]; then (echo ./pg_ctl) else (echo '') fi")
+
+##----------------------------------------------------------------------
 
 ## ---------------------------------------------------------------------
 ## Here some rules to easily manage the database.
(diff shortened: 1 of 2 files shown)


template.distillery/Makefile.db
```diff
@@ -30,8 +30,20 @@ ELIOMDEP          := $(ENV_PSQL) $(ELIOMDEP)
...
 ## ---------------------------------------------------------------------
 
+##----------------------------------------------------------------------
+
+pg_dump = pg_dump
+
+pg_ctl  = $(shell \
+           sh -c "which pg_ctl" || \
+           sh -c "which pg_ctlcluster" || \
+           sh -c "if [ -f pg_ctl ]; then (echo ./pg_ctl) else (echo '') fi")
```
vasilisp: If we end up in the empty string case, the result will be very confusing error messages, no? Maybe make up an informational non-existent executable, e.g., `cannot_find_pg_ctl`.

vasilisp: Have you tested that `pg_ctl` and `pg_ctlcluster` are compatible for the set of features you use?

dannywillems: For `pg_ctl` and `pg_ctlcluster`, see [this link](http://dba.stackexchange.com/questions/72580/missing-the-pg-ctl-package-in-postgres-9-3-installation) (see also PR #123).
The empty string is used in the conditional line 91 to say there's no pg_ctl binary found.

dannywillems: As `pg_ctlcluster` is a wrapper to `pg_ctl`, I suppose it's ok.

dannywillems: I proposed in #123 to set `pg_cltcluster` first but it's not a good idea because on Ubuntu, it doesn't work for me. The executable `pg_ctlcluster` is located in `/usr/bin/pg_ctlcluster` and it's not working.

```
$> make db-init
Makefile.mobile:242: warning: overriding recipe for target 'cordova/www/css/snapshot.css'
Makefile.mobile:156: warning: ignoring old recipe for target 'cordova/www/css/snapshot.css'
mkdir -p local_db
/usr/bin/pg_ctlcluster initdb -D local_db
Unknown option: d
Makefile.db:98: recipe for target 'db-init' failed
make: *** [db-init] Error 1
```

(just a few minutes for Debian, I got some issue with my Debian server and OPAM).

vasilisp: Debian should be just like Ubuntu in that regard.

Wouldn't it be easier and more reliable to use the `pg_ctl` in `/usr/lib/postgresql/x.y/bin/pg_ctl`? I am not sure that going through `pg_ctlcluster` is of any use to us, given that we create a local directory and do not have anything to do with the Debian/Ubuntu conventions.

dannywillems: @vasilisp : I think your idea is better than mine. I propose this command to detect the version:
```
psql --version | cut -d ' ' -f 3 | cut -d '.' -f 1-2
```

dannywillems: @vasilisp :
```Makefile
ifeq ($(shell sh -c "which psql" | echo $?),'0')
$(error "PostgreSQL is not installed")
else
PSQL_VERSION := $(shell psql --version | cut -d' ' -f 3 | cut -d'.' -f 1-2)
pg_ctl := $(shell \
           sh -c "which pg_ctl" || \
           sh -c "echo /usr/lib/postgresql/$(PSQL_VERSION)/bin/pg_ctl")
endif
```
Looks good? It supports `pg_ctl` directly (so OK for Mac OSX).

dannywillems:
@vasilisp I changed the rule for pg_ctl.

dannywillems:
@vasilisp something wrong with this PR?

vasilisp:
@dannywillems, nothing wrong, sorry, I guess the latest updates caught me in Friday mode.

PR #131, opened 2016-09-13
Add some reserve filenames for template: local, os_otdemo, db

dannywillems:
@vasilisp I forgot local and os_otdemo last time.

db is new since #123.

### DIFF
--- a/template.distillery/.reserve
+++ b/template.distillery/.reserve
@@ -1,4 +1,7 @@
+db
+local
+mobile
 options
 os
-mobile
+os_otdemo
 style


vasilisp:
What does `os_otdemo` conflict with?

dannywillems:
@vasilisp : PROJECT_NAME.css

dannywillems:
font-awesome.min is not mandatory because we can't have a dot in the project name.

vasilisp:
Thanks, you are right. Merging.

Issue #119, opened 2016-09-07
Logo Ocsigen for Ocsigen-start


dannywillems:
[See ZenHub]
For the moment, the Cordova logo is used when we install the mobile application. It will be more interesting to have the Ocsigen logo.

dannywillems:
See #132

dannywillems:
I close this issue since #132 has been merged.

PR #132, opened 2016-09-13
Ocsigen logo for icon on mobile.

dannywillems:
Issue #119 
- Makefile.mobile:
  - add icons target to generate the icons.
  - add the Ocsigen logo instead of Cordova logo by default. Use PNG instead of SVG because no good SVG Ocsigen logo found.
  - add xxhdpi and xxxhdpi for Android.
- config.xml.in:
  - Uncomment icon section for Android and iOS.
  - Add xxhdpi and xxxhdpi for Android.
- icon.png:
  - Size 512 x 512


vasilisp:
I can't find an SVG logo either. PNG will do for now. Please rebase :).

dannywillems:
@vasilisp : done, I was doing it when you sent the message :smile:.
About the SVG, I tried with the .ai but the logo is so complicated (so many different forms) that it's too complicated for me. Instead of the SVG, I generated a bigger PNG (512 x 512) which is this one.

vasilisp:
Thanks!

PR #135, opened 2016-09-14
Reorganise Makefile's and mobiles files based on BS project.

dannywillems:
- `chcp`: now we use the date and all files (when we did `make chcp`) goes into `update/%%DATE%%`.
- add splashes but not tested.

To do:
- Facebook.
- i18n
It will come in another PR.

This PR will have conflict with PR #132.

### DIFF
--- a/template.distillery/Makefile.mobile
+++ b/template.distillery/Makefile.mobile
@@ -142,7 +142,7 @@ endif
 
 $(APPJS): $(JS_PREFIX).js
 ifeq ($(APP_REMOTE),yes)
-	APPJS_FILE=$$(curl -s -f $(APP_SERVER)|grep -E -o -m 1 '$(PROJECT_NAME)_[a-f0-9]*?\.js') &&\
+	APPJS_FILE=$$(curl -s -f $(APP_SERVER) | cat | grep -E -o -m 1 '$(PROJECT_NAME)_[a-f0-9]*?\.js') &&\
 	curl -s -o $@ $(APP_SERVER)/$$APPJS_FILE
 else
 	cp -f $(WWW_PATH)/`readlink $(JS_PREFIX).js` $@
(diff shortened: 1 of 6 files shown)


template.distillery/Makefile.mobile
```diff
@@ -142,23 +142,23 @@ endif
 
 $(APPJS): $(JS_PREFIX).js
 ifeq ($(APP_REMOTE),yes)
-	APPJS_FILE=$$(curl -s -f $(APP_SERVER)|grep -E -o -m 1 '$(PROJECT_NAME)_[a-f0-9]*?\.js') &&\
+	APPJS_FILE=$$(curl -s -f $(APP_SERVER) | cat | grep -E -o -m 1 '$(PROJECT_NAME)_[a-f0-9]*?\.js') &&\
```
vasilisp: What is this `cat` for?

dannywillems:
@vasilisp : @balat asked me to merge BS Makefiles with Ocsigen-start to have something very close.
The modification for chcp comes from BS.

template.distillery/Makefile.mobile
pwbs: Haha! It's because when grep dies, it sends a SIGPIPE to curl, which doesn't like it at all. Like this, it seems to work but I agree it's a bit weird.

Instead of "cat before grep", we could use "head after grep" (as in use grep with `-m 1` and pipe to `head -n 1`) but I'm not sure it'd be much better.

vasilisp: @pwbs thanks, no need to change it. I was curious about what crazy glitch necessitates it :). Maybe add a comment when you get a chance, because I am sure somebody will propose removing the useless `cat`.

pwbs: @vasilisp Yup, it'd be better with some comments! @dannywillems :p

PR #138, opened 2016-09-14
Prefix tables with os_

dannywillems:
I can't test it because the template is in progress but it compiles...

### DIFF
--- a/src/os_db.ml
+++ b/src/os_db.ml
@@ -116,11 +116,11 @@ end
 
 (*****************************************************************************)
 (* tables, for Macaque *)
-let users_userid_seq = <:sequence< bigserial "users_userid_seq" >>
+let os_users_userid_seq = <:sequence< bigserial "os_users_userid_seq" >>
 
-let users_table =
-  <:table< users (
-       userid bigint NOT NULL DEFAULT(nextval $users_userid_seq$),
+let os_users_table =
+  <:table< os_users (
+       userid bigint NOT NULL DEFAULT(nextval $os_users_userid_seq$),
        firstname text NOT NULL,
        lastname text NOT NULL,
        main_email citext NOT NULL,
@@ -128,18 +128,18 @@ let users_table =
        avatar text
           ) >>
 
-let emails_table =
-  <:table< emails (
+let os_emails_table =
+  <:table< os_emails (
        email citext NOT NULL,
        userid bigint NOT NULL,
        validated boolean NOT NULL DEFAULT(false)
           ) >>
 
-let action_link_table :
+let os_action_link_table :
   (< .. >,
    < creationdate : < nul : Sql.non_nullable; .. > Sql.t > Sql.writable)
     Sql.view =
-  <:table< activation (
+  <:table< os_activation (
        activationkey text NOT NULL,
        userid bigint NOT NULL,
        email citext NOT NULL,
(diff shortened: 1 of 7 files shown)


sagotch:
I guess we also need some kind of `ALTER TABLE name RENAME TO os_name` to update existing database name, otherwise os_* stuff will not work anymore.

dannywillems:
@sagotch Good point. I think it's related to what we talked yesterday: we need to have SQL files to migrate from one version to another.

vasilisp:
So, should we wait for a working migration script? Or are you planning to do an one-off manual migration of BS?

dannywillems:
@balat ?
I'm not sure if there is a need to do a working migration script now as there's only BS which uses OS. 

But first we need to define a structure for the migration scripts (do we download it separately? Or it is in the template? Do we give a migration script for version 1 to 3, 1 to 2 and 2 to 3 or only for version 1 to 2 and version 2 to 3?, etc).
It's not very clear because we talk about it only 5 or 10 minutes.

balat:
I think we can do a manual migration of BS, as we are probably the only one concerned (I think).
Let's wait to have this before merging this PR.

sagotch:
FWIW, I have other projects using ocsigen-start. Not very active but I will eventually need to update them with new db scheme.

dannywillems:
I think these lines are enough for the moment (references seem to be updated automatically):
```SQL
ALTER TABLE activation  RENAME TO os_activation;
ALTER TABLE emails      RENAME TO os_emails;
ALTER TABLE groups      RENAME TO os_groups;
ALTER TABLE preregister RENAME TO os_preregister;
ALTER TABLE user_groups RENAME TO os_user_groups;
ALTER TABLE users       RENAME TO os_users;
```
I think `update.sql` is a good name but where do I put this file? It doesn't make sense to put in the template because if someone began with it, he won't never recreate the template.

sagotch:
It would be safer to check if `os_*` tables already exist before trying to rename it.

If the user want to create a table `users`, running the update twice will not be a problem if os table name do not change because the `alter table` will fail, but if for instance `os_users` become `os_users_trololo`, the update will finally happen. So before changing unprefixed name, we should look for prefixed version, just in case.

dannywillems:
As copying the file `update.sql`in the template doesn't make sense (because if someone creates a new template, it will come with the more recent `PROJECT_NAME.sql`and `update.sql`is useless), I implemented a way to ignore some files while copying in `eliom-distillery` (see [this PR](https://github.com/ocsigen/eliom/pull/374)).

I suggest to add a rule `make db-upgrade` in `Makefile.db` which will call the `update.sql`in the template OPAM directory. @vasilisp What do you think?

pwbs:
> I suggest to add a rule make db-upgrade in Makefile.db which will call the update.sqlin the template OPAM directory

Is it just me or it's a bit weird to have a rule named db-*upgrade* call something called *update*?

vasilisp:
@dannywillems I agree on a `db-upgrade`.

We do need to fix the `db-update` (re: @pwbs). It is more like initialization, but we already have a `db-init` that initializes the database *daemon*, as opposed to `db-update` that initializes the thing that Postgres calls a *database* (terminology overload). Maybe `db-update -> db-schema`?

vasilisp:
I was actually confused above about what update we are talking about, but that actually reinforces the point about weirdness :). So, `make db-schema` and `upgrade.sql` for the table upgrades?

dannywillems:
@vasilisp : Agree on `db-schema`.
So:
- `db-update -> db-schema`
- `db-upgrade` calls `upgrade.sql`in the template OPAM directory to upgrade the database schema to the newer OS?

dannywillems:
@vasilisp : I updated. The `upgrade.sql` comes from the template OPAM directory.
It's related to [this PR in eliom](https://github.com/ocsigen/eliom/pull/374).

vasilisp:
The `upgrade.sql` solution looks OK for now, but we may want a more sophisticated (basically versioned) approach in the feature; and we need to be very prudent about schema changes moving forward.

@balat, looks ready to me, so you can merge it when the BS migration is done.

dannywillems:
@balat ?

balat:
@vouillon suggest to use SQL's schemas to group table together, and write ocsigenstart.users instead of os_users.

dannywillems:
@balat `ocsigen_start.users` instead of `ocsigenstart.users`

dannywillems:
Done.

vouillon:
For migration, one can first create the schema and then set the search path:
```
CREATE SCHEMA `ocsigen_start`;
ALTER DATABASE bs SET search_path TO ocisgen_start,public;
```
All connections to the database then need to be restarted to get the right search path.
Finally, we can move the table to the `ocsigen_start` schema:
```
ALTER TABLE users SET SCHEMA ocsigen_start;
```

PR #140, opened 2016-09-15
create ocsipersist database on "make db-create"

jrochel:


dannywillems:
@jrochel It's not created in the SQL file?

jrochel:
That's right, but I thought that it's better to include it here too as it's a CREATE DATABASE, not just a CREATE TABLE, and thus on a different level. To be honest I'm not sure whether this is actually very helpful but I thought it couldn't hurt.

dannywillems:
@jrochel Maybe add `IF NOT EXISTS` to avoid error when we have multiple ocsigen-start projects?
It's also `CREATE DATABASE ocsipersist` in `PROJECT_NAME.sql`.

jrochel:
"CREATE DATABASE IF NOT EXISTS" NOT EXISTS.
I could add a "|| true" in the Makefile but I don't really see the point. But I have removed the line from PROJECT_NAME.sql.

Issue #56, opened 2016-04-13
CSS inclusion order matters


sagotch:
Current way to include css is buggy, since rule definition override rules defined earlier.

Manually defining sheets order in the only way to go.

dannywillems:
@sagotch Is it resolved?

sagotch:
The old way to include css file no longer exist. As we are back to manual definition of order of sheets, I close the issue.

Code involved was:
```
let%server include_files_from dir =
  let css_files = Sys.readdir dir in
  Array.to_list @@ Array.map (fun x -> [x]) css_files
```
in `PROJECT_NAME_page.eliom` file.

PR #147, opened 2016-09-20
Makefile style.

dannywillems:
### For the moment:
- Add comments to explain the Makefile and change some variable names to be more explicit.

### To-do:
Some work must be also done in `PROJECT_NAME_page.eliom`. CSS files must be concatenated in only one because on iOS, the application blinks when changing page. This blink issue is resolved by adding `let%client _ = Eliom_client.persist_document_head ()` but it forces to add all CSS files in the header of `eliom.html.in` which is not very convenient.
A solution is to concatenate all CSS files in `PROJECT_NAME.css` (by the way, this file isn't included in the website because it's not listed in `the_local_css` and in the mobile version it's replaced) but it must be concatenated in order. The wildcard used for the moment isn't a good solution because we have the same issue than #56.
Here some ideas:
- instead of using a wildcard, we write the filenames in the variable `EXTERNAL_CSS_FILES`.
- use SASS and the dev uses `@import 'filename'` (or not if he doesn't care about the order or if he has only one SASS file).
 
@vasilisp @balat: Do we force devs to use SASS? A disadvantage is the dev must install Ruby before to install SASS. But SASS seems to be widely used so it's not very inconvenient.
I wait your decision to continue.


vasilisp:
I would highly appreciate it if the SASS functionality stayed optional, e.g., in the form of inactive or commented-out makefile rules. SASS/CSS in the periphery of what we are offering, which is a skeleton for a multi-user app. CSS suffices for the default style, which has to stay lean anyway.

I do think it would be annoying to drag-in another language implementation. All the Node/Java things we need for mobile development are enough of a pain already :).

balat:
It is difficult to make SASS optional, as we would have to provide and maintain both.
I don't know what is the best:

1. SASS is difficult to install (and this adds a dependency)
2. but Ocsigen Start is supposed to make (reasonable) choices for the user, to avoid having to reinvent the wheel. And I would really recommend to use SASS instead of basic CSS.

I think the answer is in the amount of difficulty of 1. Is it really a nightmare? I don't remember.

dannywillems:
@balat about SASS installation, It's quite easy. I did it yesterday (I hope I forget nothing).

- Install RVM (Ruby Version Manager): `curl -L https://get.rvm.io | bash -s stable --autolibs=enabled`.
- Add RVM bin dir to PATH.
- Install a ruby version with `rvm install [version]`.
- Install SASS with `gem install sass`.

Drup:
I also think SASS should be optional and, possibly more importantly, not necessary for deployment.

Why not just have the SASS version, have a phony makefile rule (`make sass`) that generates CSS from it, and include the CSS pre-generated by default ?

balat:
The most important thing for me is that Ocsigen Start should be very easy to install and use ("your app in 5 minutes").
If I'm right, there is a SASS package in Ubuntu and Debian?
Which solves the problem in my opinion. OS depends on postgresql and sass (amongst many other dependencies).

The goal of Ocsigen Start is to provide a ready to use application, and save the developers the need of doing all the choices we took time to make for xprime and besport. We want to help developers to concentrate on the core of their app (and not implementing connection forms, activation links, choosing a database, or choosing whether to use css, sass or less ...). The goal is to make it possible to create apps MVPs very quickly, which is necessary if we want to see more Ocsigen apps.

Obviously, it must always be possible to to a different choice (switch to another db or to less, remove some part of the default app). In the future, Ocsigen Start should be completely configurable.

dannywillems:
> If I'm right, there is a SASS package in Ubuntu and Debian?

Yes, it's `ruby-sass`.
On Mac OS X, ruby and gem are installed by default so `sudo gem install sass` does the job.

dannywillems:
My last commit satisfies everyone:
- SASS is optional. A variable `USE_SASS` is defined in `Makefile.options` which allows to choose. Set to another value of `yes` deactivate SASS.
- SASS is activated by default.
- The SASS directory is `sass` in the root directory.
- When uploading the style, the CSS files `static/css/os.css` and `static/css/os_otdemo.css` **must be generated** from `sass/os.scss` and `sass/os_otdemo.scss`. In this way, the CSS is always up-to-date with the equivalent SCSS.
- All CSS files are concatenated in only one file `$(LOCAL_CSS)` which is by default `static/css/PROJECT_NAME.css` (see why in my first message).
- It's mandatory to list all OT files one-by-one for the stylesheet order.
- And I documented everything. So for information and details, read the comments.

PR #149, opened 2016-09-20
Icons

dannywillems:
- Remove Ow_icons dependency.
- Add comments.
- ~~I also removed the `class_` parameter because it seems to do the same job than classes.~~

I didn't merge `Ot_icons` and `PROJECT_NAME_icons.eliom` and I didn't include `Ot_icons` in `PROJECT_NAME_icons.eliom` because they define two different families of icons: one with OT, the other with Font Awesome.

### DIFF
--- a/template.distillery/PROJECT_NAME_icons.eliom
+++ b/template.distillery/PROJECT_NAME_icons.eliom
@@ -1,24 +1,37 @@
+(** This module defines an interface to create icons HTML element with
+ * predefined style/value. It is supposed "Font Awesome" icons are used by
+ * default (fa CSS class is added when using [icon classes]).
+ * See http://fontawesome.io/ for more information and for the complete list of
+ * CSS classes values.
+ *)
+
 [%%shared.start]
 
 module Make(A : module type of Eliom_content.Html.F) = struct
 
-  let icon classes ?(class_=[]) () =
-    A.i ~a:[A.a_class ("fa"::classes@class_)] []
+  (** [icon classes] create an icon HTML attribute with "fa" and [classes]
+   * as CSS classes.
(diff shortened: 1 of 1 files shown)


template.distillery/PROJECT_NAME_icons.eliom
```diff
@@ -1,24 +1,37 @@
...
+  (* Example for the user icon:
+   *  let user = icon ["fa-user"; "fa-fw"]
+   *)
 
 end
 
 module F = struct
-  include Ow_icons.F
   include Make(Eliom_content.Html.F)
 end
```
vasilisp: It can become simply `module F = Make(Eliom_content.Html.F)`, and same for `D`.

template.distillery/PROJECT_NAME_icons.eliom
```diff
@@ -1,24 +1,37 @@
...
-  let icon classes ?(class_=[]) () =
-    A.i ~a:[A.a_class ("fa"::classes@class_)] []
+  (** [icon classes] create an icon HTML attribute with "fa" and [classes]
+   * as CSS classes.
+   * The optional parameter is at the end to be able to add other CSS classes
+   * with predefined icons.
+   *)
+  let icon classes
+      ?(a = ([] : Html_types.i_attrib Eliom_content.Html.attrib list)) () =
+    A.i ~a:(A.a_class ("fa" :: classes) :: a) []
```
vasilisp: `<i>` has traditionally meant italics. It is news to me that it is now widely used for icons :). It is OK if we do the same, but can you clarify that it is an `<i>` tag in the documentation comment?

dannywillems:
@vasilisp done. I also added more explanation about the i tag and the optional parameter.

PR #155, opened 2016-09-22
Remove section in OS and in the template.

dannywillems:
- Remove section ([%%server ...], ...) and replace it with let ([let%server, ...]).
- Analyze the convention about spaces for [%%server ...] (for cases which is not in the top level) which is two spaces and propagate it to all files.

Normally, everything has been changed.

Asked by @balat.


dannywillems:
Conflicts fixed.

vasilisp:
@dannywillems looks OK, but I suspect it conflicts with other PRs, so let me know in what order you would prefer to merge things. (Feel free to merge it yourself if you have the appropriate permissions.)

vasilisp:
As a meta-comment, I personally prefer to do these kinds of changes only at the same time as "real" changes to the respective files, because otherwise we kill `git blame` and have to deal with huge hard-to-review diffs. But since you did it, we can merge it.

dannywillems:
@vasilisp About permissions to merge, I prefer to have a code review before (even if it's a small change). @balat gave me rights for ZenHub and so I'm allowed to assign me tasks.

@vasilisp Ok for the meta-comment.

dannywillems:
@vasilisp About conflicts, I will resolve them if any. This PR comes before other to have something clearer.

PR #152, opened 2016-09-22
Comments in Os lib.

dannywillems:
*I open this PR now because it will be a big PR so you can see modifications step by step.*

I began to comments all files in Os ~~and in the template.~~
It's useful to know what it is really used, what is deprecated, what can be removed, etc. It's also useful for people who will use the template and Os (to improve it or to use it).

I don't know what is the convention for documentation (I saw several styles) so don't hesitate to tell me if something is not well done (mostly on my English).

When I will finish the entire documentation, it will be in only one commit. For the moment, I do separate commit for more visibility and to follow more easily.

PR #158, opened 2016-09-23
Interfaces for library.

dannywillems:
This PR add missing interfaces to `eliom` and `ml` files ~~in the template and~~ in the library.

It aims to simplify the use of the library and the template by providing more clearer description of functions and modules within the interfaces.

Each commit is for a different interface.

Comments in these interfaces will be done in the PR #152 so this PR must be merged first.


dannywillems:
I think it can be merged. Interfaces in the template will be in another PR.

vasilisp:
The code looks OK. It is impossible to mentally type-infer everything, so let's merge and we will notice if something is too strict later :).

dannywillems:
@vasilisp I used `eliomc -i` to infer and I added what was missing (and resolved bad inference with server/client/shared sections.

PR #160, opened 2016-09-23
Add interfaces to template.

dannywillems:
I don't like the `PROJECT_NAME_page.eliomi` due to the `include` in the `eliom` file, but I have no idea how to bypass it.

`PROJECT_NAME_base.eliomi` isn't done.
`PROJECT_NAME_content.eliomi` isn't also done because some things will be removed in next PR.


dannywillems:
Can someone review this PR please? I need it to continue #152.

template.distillery/PROJECT_NAME_container.eliomi
```diff
@@ -0,0 +1,34 @@
...
+
+[%%server.start]
+
+val get_wrong_pdata :
+  unit -> ((string * string) * (string * string)) option Lwt.t
+
+[%%client.start]
+
+val get_wrong_pdata :
+  (Deriving_Json.Json_unit.a, ((string * string) * (string * string)) option)
```
vasilisp: Can you remove the confusing `Deriving_json` paths from the types? For example, I think this is just `unit`.

template.distillery/PROJECT_NAME_handlers.eliomi
```diff
@@ -0,0 +1,59 @@
+(* This file was generated by Ocsigen-start.
+   Feel free to use it, modify it, and redistribute it as you wish. *)
+
```
vasilisp: Is this a superset of the `Os_handlers` signature? If yes, it would be much more maintainable if we had an `include module type of Os_handlers`.

template.distillery/PROJECT_NAME_container.eliomi
dannywillems: Fixed.

template.distillery/PROJECT_NAME_handlers.eliomi
dannywillems: I rearranged `PROJECT_NAME_handlers.eliom[i]` because there were duplicated handlers due to the old `include Os_handlers`.
I also changed the signature of the client handlers by removing `Deriving_Json` as in the previous comment.

dannywillems: I also added some comments to separate the different handlers because it was confusing to see no separation between the handlers definitions.

dannywillems:
@vasilisp I fixed what you pointed out yesterday. Is it ok for you?

src/os_handlers.eliom
```diff
@@ -25,6 +25,7 @@
   open Eliom_content.Html.F
 ]
 
+(* Set personal data *)
```
vasilisp: When we have a good name that accurately describes the functionality, there is no need to duplicate the description as a comment :). But no need to remove these now.

PR #166, opened 2016-09-27
Userid --> myid convention

dannywillems:


vasilisp:
To be honest, I find `user` clearer. Does `my` resolve some ambiguity somewhere?

dannywillems:
@balat said the convention was `myid` for functions called in a session. For me, it doesn't make a big difference. I even prefer `userid` than `myid`.

vasilisp:
OK, makes sense if we are consistent about it. Please rebase :).

dannywillems:
@vasilisp I wait the merge #167 because there will be new conflicts...

dannywillems:
@vasilisp Ok. Fixed.

PR #167, opened 2016-09-27
Remove simple quote at the end of some handlers and services.

dannywillems:
I suppose it was the previous convention naming (before the changes about services declarations).

### DIFF
--- a/src/os_db.mli
+++ b/src/os_db.mli
@@ -22,6 +22,7 @@
 exception No_such_resource
 exception Main_email_removal_attempt
 exception Account_already_activated
+exception Account_not_activated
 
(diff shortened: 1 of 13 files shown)


vasilisp:
I am very happy about removing these :). @balat you will need to update against this. Do we merge now or later?

dannywillems:
It remains some functions defined with `'` at the end but it's to avoid conflicts with the function without `'`. I don't know if it's a good convention. It's error prone.

dannywillems:
@vasilisp Rebased. I also added `exception Account_not_activated`in `os_db.mli` which comes from another PR.

PR #168, opened 2016-09-27
Activation links now called Action links + make it possible to customize handler

balat:


vasilisp:
Is this superseded by #177 ?

PR #171, opened 2016-09-28
specify DB connection pool size

jrochel:


src/os_db.ml
```diff
@@ -60,12 +59,23 @@ let connect () = Lwt_PGOCaml.connect
...
 
-let validate db =
-  try_lwt
-    lwt () = Lwt_PGOCaml.ping db in
-    Lwt.return true
-  with _ ->
-    Lwt.return false
+let pool_r = ref @@ Lwt_pool.create 16 ~validate connect
+
+let pool : (string, bool) Hashtbl.t Lwt_PGOCaml.t Lwt_pool.t = !pool_r
```
vouillon: This does not work. We want to always use the most recent pool, but [pool] refers to the initial one.

src/os_db.ml
```diff
@@ -43,13 +43,12 @@ let password_r = ref None
...
 let unix_domain_socket_dir_r = ref None
 
-let init ?host ?port ?user ?password ?database ?unix_domain_socket_dir () =
-  host_r := host;
-  port_r := port;
-  user_r := user;
-  password_r := password;
-  database_r := database;
-  unix_domain_socket_dir_r := unix_domain_socket_dir
+let validate db =
```
vouillon: You don't have to move the [validate] function.

src/os_db.ml
jrochel: I will never get it into my head... functional programming without lazy evaluation...

PR #176, opened 2016-09-29
Os_view - check password function --> extract it.

dannywillems:
It avoids to duplicate code and it allows to use it independently of defined forms.

I think this kind of function must be in a submodule or another file. I think forms in `Os_view` must be in a submodule `Os_view.Forms`.

### DIFF
--- a/src/os_view.eliomi
+++ b/src/os_view.eliomi
@@ -19,6 +19,17 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
 
+[%%client.start]
+
+(** [check_password_confirmation ~password ~confirmation] adds a listener to
+    the element [confirmation] which checks if the value of [password] and
+    [confirmation] match.
+ *)
+val check_password_confirmation :
+  password:[< Html_types.input] Eliom_content.Html.elt ->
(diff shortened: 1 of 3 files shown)


src/os_view.eliomi
```diff
@@ -19,6 +19,17 @@
...
 
+[%%client.start]
+
+(** [check_password_confirmation ~password ~confirmation] adds a listener to
+    the element [confirmation] which checks if the value of [password] and
+    [confirmation] match.
+ *)
+val check_password_confirmation :
+  password:[< Html_types.input] Eliom_content.Html.To_dom.elt ->
+  confirmation:[< Html_types.input] Eliom_content.Html.To_dom.elt ->
```
Drup: `To_dom.elt` ?

dannywillems: @Drup Fixed. I removed the `To_dom`. It was the type given by the type inference.

vasilisp:
> I think forms in `Os_view` must be in a submodule `Os_view.Forms`.

If you do start one, please call it `Form`. Our convention is to use singular for such things.

dannywillems:
@vasilisp Oh I thought it was plural because in `PROJECT_NAME_content.eliom`, it's `Forms`. I will use the singular convention.

PR #184, opened 2016-10-03
Os_types

dannywillems:
Instead of using `int64` in `Os_db.mli` as a user ID, it's better to use something like `Os_types.userid`. With this method, `Os_db.mli` can be compiled independently from `Os_user.eliomi`, we don't have compilation loops and we have a clearer interface.
In the future, any ID or any types used in many modules can be put in `os_types.eliom` for the same reason.

So I moved:
- `Os_user.id` --> `Os_types.userid`
- `Os_user.t` --> `Os_types.user`
- `Os_group.id` --> `Os_types.groupid`
- `Os_group.t` --> `Os_types.group`

I'm not sure the second and the fourth are useful (because the types are not abstract) so I did a different commit if the majority thinks `Os_user.t` and `Os_group.t` are better.

There's no `eliomi` file because it would be the same.

### DIFF
--- a/src/os_current_user.eliom
+++ b/src/os_current_user.eliom
@@ -24,7 +24,7 @@ let section = Lwt_log.Section.make "os:current_user"
   type current_user =
     | CU_idontknown
     | CU_notconnected
-    | CU_user of Os_user.t
+    | CU_user of Os_types.user
 ]
 
 let%shared please_use_connected_fun =
--- a/src/os_current_user.eliomi
+++ b/src/os_current_user.eliomi
@@ -23,11 +23,11 @@
 type current_user =
   | CU_idontknown
   | CU_notconnected
-  | CU_user of Os_user.t
+  | CU_user of Os_types.user
 
-(** [get_current_user ()] returns the current user as a {!Os_user.t} type.
+(** [get_current_user ()] returns the current user as a {!Os_types.user} type.
     If no user is connected, it fails with {!Os_session.Not_connected}. *)
-val get_current_user : unit -> Os_user.t
+val get_current_user : unit -> Os_types.user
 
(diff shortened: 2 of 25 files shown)


vasilisp:
@dannywillems, I don't like the new scheme. We will sooner or later need utilities for all these types. Are we going to put them in `Os_types` (along with the type, as one would expect), or in `Os_user` & co. (as one would expect from the subject matter)? `Os_types` would grow to a hodge-podge à la `Eliom_common` :).

dannywillems:
@vasilisp `Os_types` is only to define types. No utilities for these types must be written in `Os_types`. So to answer, it would be in `Os_user` & co. Maybe I must write it in comments?

The main reasons of this separation is, as described:
- avoid loop compilations.
- clearer interfaces.

This scheme (`userid` and `groupid`, not sure about `user`) is used in BS for reasons I described.

Drup:
Why not also keep the type aliases in the various modules, so that you can also say `Os_user.t` (and so on) ?

dannywillems:
@Drup I don't think duplicate types is a good idea.

Drup:
It's not duplicating, it's re-exporting. In `Os_user`: 

```ocaml
type id = Os_types.id

type t = Os_types.user = {
    userid : id;
    fn : string;
    ln : string;
    avatar : string option;
  } [@@deriving json]
```

Duplication is a problem if things get out of sync. That can't happen here.

dannywillems:
I'm not sure if it is useful. I let other to give their suggestions.

dannywillems:
@Drup I add the type aliases.

dannywillems:
Rebased with the last PR #197. I also replaced tab by spaces introduced in this last PR.

PR #185, opened 2016-10-03
Os_db: export full_transaction_block function and add similar functions

vouillon:


vasilisp:
@vouillon what is the rationale behind having both `without_transaction` and `single_request`? Are the implementations going to diverge in the future? Given that we do not impose a single request at the type level, it seems to me that we can simply document the fact that doing multiple is risky business.

vouillon:
I have removed `single_request`.

vasilisp:
Let's get another PR out of the way then :).

PR #186, opened 2016-10-03
Rename back Os_user.update to Os_user.update'

vouillon:
As it shadows existing OS_user.update function...


vasilisp:
I think we can merge this. @dannywillems, feel free to propose a better name.

dannywillems:
@vasilisp I would like to change this `include Os_db.User` in `Os_user.eliom` to have a clearer interface. I'm on it but it's not easy for naming.

PR #188, opened 2016-10-03
Userid as first parameter for consistency.

dannywillems:
If it's ok, BS must be changed first @balat.

### DIFF
--- a/src/os_db.ml
+++ b/src/os_db.ml
@@ -358,7 +358,7 @@ module User = struct
        d.userid = $int64:userid$
       >>
 
-  let update_password userid password =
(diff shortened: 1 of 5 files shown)


vasilisp:
Looks OK to me.

pwbs:
Since we're changing that... why not label the parameters?

dannywillems:
@pwbs You will also need to change BS because some functions aren't labeled and the order changed.

pwbs:
> @pwbs You will also need to change BS because some functions aren't labeled and the order changed.

Yes, of course. My point is, since we're breaking the compatibility, we may make even more changes. :)

PR #189, opened 2016-10-03
Remove 'empty' user from Os_user

dannywillems:
I don't know what it is used for. It's not exported in the interface so I supposed it was only for internal uses but nothing is using it in  `Os_db`. Any explanation? Instead of an issue, I propose the PR directly if there is no reason.

### DIFF
--- a/src/os_user.eliom
+++ b/src/os_user.eliom
@@ -113,13 +113,6 @@ let password_set userid =
 
 *)
 
-let empty = {
-  userid = 0L;
-  fn = "";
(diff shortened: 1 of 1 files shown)


dannywillems:
@vasilisp Do you know why there were this empty value?

vasilisp:
Probably needed at some point, but I don't feel motivated to do `git blame` gymnastics and find out why :). Software evolves, and we are not always careful about eliminating code that becomes dead.

PR #191, opened 2016-10-03
get_email_validated --> is_email_validated

dannywillems:
I think it makes more sense.

**Since it's changed in the template, I suppose it must be changed in BS.**


Drup:
`is_email_validated` ? 
`validated_email_is` ? 
:yoda:

dannywillems:
`is_email_validated` is better I think.

dannywillems:
@sagotch @pwbs @balat @vouillon @jrochel. Could it be merged?

PR #196, opened 2016-10-04
simplified notification interface

jrochel:
- split notification type into a server type and a client type
- don't display notification for your own chat messages (works now also for multi-server set-up)


vouillon:
I will just rebase the `bs` branch onto master. This is simpler than maintaining two ever diverging branches.

PR #199, opened 2016-10-06
Push notifications: GCM --> FCM and parse the response.

dannywillems:
It wasn't relevant to implement it before now.

I also add `string_of_stream` in the module `Os_lib` to read a response of an HTTP post request. It comes from bs.

I'm trying to test all error returned by GCM because the Google documentation is not well done. It works for the moment but I would like to test all case.

An important thing about the review it's to check if I always finalize the stream ( @vouillon ).

### DIFF
--- a/src/os_lib.eliom
+++ b/src/os_lib.eliom
@@ -30,3 +30,18 @@ let%shared memoizator f =
       let%lwt value = f () in
       value_ref := Some value;
       Lwt.return value
+
+[%%server.start]
+module Http =
+  struct
+    let string_of_stream ?(len=16384) contents =
+      Lwt.try_bind
+        (fun () ->
+           Ocsigen_stream.string_of_stream len (Ocsigen_stream.get contents))
+        (fun r ->
+           let%lwt () = Ocsigen_stream.finalize contents `Success in
+           Lwt.return r)
+        (fun e ->
+           let%lwt () = Ocsigen_stream.finalize contents `Failure in
+           Lwt.fail e)
+  end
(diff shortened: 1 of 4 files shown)


dannywillems:
I can't reproduce all errors sent by FCM (timeout, internal server error, etc) so I can test if the doc is up-to-date. I parse errors listed [here](https://developers.google.com/cloud-messaging/http-server-ref#table9).

dannywillems:
Since the Google migration to FCM, the interface must be changed and also the Cordova plugin (phonegap-plugin-push is not yet compatible with FCM and branch about FCM compatibility doesn't seem to be up-to-date).
I suggest to switch to cordova-plugin-fcm like some other people did. I did the binding (https://github.com/dannywillems/ocaml-cordova-plugin-fcm) and it's working on iOS and Android (tested).

dannywillems:
So I tested:
* [x] Retro-compatibility with PhoneGap (only Android, never used on iOS and phonegap-plugin-push is still not working for me...
* [x] Payload defined by FCM. Ok for iOS 9 and Android. Can be found [here](https://github.com/dannywillems/ocsigen-mobile-push-notification-example/tree/fcm). @venator will also test it for bs.
* Error messages.
  * [x] Mismatch sender ID
  * [x] Missing registration
  * [x] Invalid registration
  * [x] Unregistered device
  * [ ] ...?

I think it could be merged when BS will have implemented push notifications (WIP).

venator:
Ok for me.

PR #200, opened 2016-10-06
Update phonegap-plugin-push to 1.8.2

dannywillems:
I tested it *on Android*. It's OK.


vasilisp:
Possibly stupid question, but does this one also need testing on iOS?

dannywillems:
@vasilisp I never succeded on iOS when GCM was used because I didn't understand the interface. #199 switches to FCM and I will try it in the next few days.
As I never tested on iOS, it can't be worst than now. And I think it's better to update in any case.

PR #201, opened 2016-10-06
Update hot code push to 1.5.1

dannywillems:
Ok on Android.
See [changelog](https://github.com/nordnet/cordova-hot-code-push/blob/master/CHANGELOG.md)

@pwbs Could you test it on iOS?


dannywillems:
Ok on iOS with a simple project.

vasilisp:
OK, let's merge. It is an easy-enough downgrade if it turns out to be problematic.

PR #208, opened 2016-10-07
Empty .depend file in template and update .gitignore.

dannywillems:
- `.depend` is added in the template to avoid the warning `Makefile.os:275: .depend: No such file or directory`
- about `.gitignore`, I suppose it was old rules. Never this kind of files would be put in `template.distillery` directory now we use `eliom-distillery`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,15 +21,4 @@ local/
 META
 Makefile.depend
 
-samples/static/avatars
-samples/static/tmp
-
-# local database directory
-samples/data
-samples/local
-
-samples.distillery/local*
-samples.distillery/_*
-samples.distillery.*
-
 _opam
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@ _client/
 _deps/
 _server/
 local/
+doc/client
+doc/server
 
 .*.sw*
 


vasilisp:
Are you sure that the empty `.depend` is OK? Can it be that we don't generate a proper `.depend` because there is a new-enough empty one? Haven't examined it in depth, just speculating.

dannywillems:
@vasilisp I haven't examined it in depth also but when I tried to compile the template it was ok. I don't know much. Maybe wait for someone who knows more about the subject?

vasilisp:
Given that it is only a minor annoyance and it hasn't caused us any trouble, we can merge just the first commit and think about improving the dependency management later. Force-push just the first commit if you agree :).

dannywillems:
@vasilisp Ok done. Do you think it is useful to add `doc/client` and `doc/server` in `.gitignore`? These directories are created by `make doc`.

vasilisp:
@dannywillems, sure. Our `git status` should be free of auto-generated files.

dannywillems:
@vasilisp Ok, added.

vasilisp:
@dannywillems didn't you mean to add some rules about `doc`? The current version only removes some :).

dannywillems:
@vasilisp Hum. I don't know what I did... It's OK now.

PR #212, opened 2016-10-07
Makefile.mobile: Add icon-20 and icon-20@2x for iOS

dannywillems:
Other icons are missing but I can see it for the moment (need to update Xcode I think...).


dannywillems:
@vasilisp By the way, CI for PR is not fixed.

PR #213, opened 2016-10-10
Ios Icons generation: change order in Makefile.

dannywillems:
because the first `icon-%` was used instead of `icon-%2x`.

It caused bad icons generation.


dannywillems:
This issue was resolved in GNU Make 3.82 (2010) (so this PR is not relevant for people who uses this version or later) but by default, on Mac OS X, XCode installs GNU Make 3.81 (2006...)

PR #232, opened 2016-10-19
Hotfix ocsigen-toolkit dependency in eliomdoc and remove warnings from eliomdoc.

dannywillems:
- Fix documentation generation by adding ocsigen-toolkit as dependency in eliomdoc packages
- Hide warnings to avoid to see all `...not found...`. I don't know if it's a good solution because it hides also other warnings, but which other warnings could we have?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -219,21 +219,23 @@ COMMON_OPTIONS := -colorize-code -stars -sort
 eliomdoc_wiki = ODOC_WIKI_SUBPROJECT="$(1)" \
                 eliomdoc \
                 -$(1) \
-                -ppx -package pgocaml,yojson,calendar \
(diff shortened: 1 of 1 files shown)


vasilisp:
Does this restore `make doc` on your machine? I tried to test locally before committing an identical fix, but I was getting errors about an unrecognized extension (`.cma`) and I gave up.

dannywillems:
@vasilisp I suppose you're on 4.03.0. Wikidoc is only available for 4.02.3 so you need to switch on this compiler, install ocsigen and Wikidoc.

dannywillems:
@vasilisp is it OK now to generate documentation?

vasilisp:
@dannywillems, I haven't tried locally. But it works in Jenkins, so I am happy enough :).

PR #238, opened 2016-10-20
fix: security issue in account creation (regarding emails)

pwbs:
modified:   src/os_db.ml

### DIFF
--- a/src/os_handlers.eliom
+++ b/src/os_handlers.eliom
@@ -295,7 +295,7 @@ let action_link_handler_common akey =
       else Lwt.return_unit
     in
     let%lwt () =
-      if action = `AccountActivation
+      if action = `AccountActivation || action = `PasswordReset
       then Os_db.User.set_email_validated userid email
       else Lwt.return_unit
     in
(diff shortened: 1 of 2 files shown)


dannywillems:
@pwbs I think it's interesting to export the function `pg_escape_quotes` for other PGOCaml requests.

pwbs:
`pg_escape_quotes` has disappeared... ^^' 

Do you want it to be somewhere?

dannywillems:
**!! Don't merge: it needs a database connection to compile. !!**

pwbs:
Indeed, I have to rewrite that using macaque :( :( :( :( :( :(

src/os_handlers.eliom
```diff
@@ -295,7 +295,7 @@ let action_link_handler_common akey =
       else Lwt.return_unit
     in
     let%lwt () =
-      if action = `AccountActivation
+      if action = `AccountActivation || action = `PasswordReset
```
pwbs: (preventing registering an email that's already been through the registration process) => (validate emails for which we require a password reset)

pwbs:
I've push-f-ed and it's now using macaque...

It's incredibly difficult to use macaque...

PR #245, opened 2016-10-25
Remove %_db2 target. Useless

dannywillems:


vasilisp:
I think there is a `_db2` module in BeSport and this target was left there for consistency? But anyway, it is indeed useless for OS, so let's remove it.

PR #248, opened 2016-10-26
Add ImageMagick dependency for convert utility

dannywillems:


README.md
```diff
@@ -43,12 +43,14 @@ opam pin add --no-action -y ocsigen-start https://github.com/ocsigen/ocsigen-sta
...
 ```
 
-Ocsigen-toolkit depends on the [OCaml binding to ImageMagick C library](https://github.com/besport/ocaml-imagemagick) so you need to install the [C library](http://www.imagemagick.org/script/index.php).
+Ocsigen-start depends on the [OCaml binding to ImageMagick C library](https://github.com/besport/ocaml-imagemagick) so you need to install the [C library](http://www.imagemagick.org/script/index.php).
 
 On Mac OSX, you can install it with Homebrew: `brew install imagemagick`.
 
 On Debian/Ubuntu , you can install it with apt-get: `sudo apt-get install libgraphicsmagick1-dev libmagickcore-dev`
 
+`Makefile.mobile` needs `convert` which is an ImageMagick utilities. If you are on Debian/Ubuntu, you can install these utilities with `sudo apt-get install imagemagick`.
```
vasilisp: There is some singular/plural mismatch here :).

dannywillems:
@vasilisp Fixed.

PR #253, opened 2016-10-26
change behavior for already used activation keys

pwbs:
edited:
# was

modified:   src/os_db.ml
modified:   src/os_db.mli
modified:   src/os_handlers.eliom
modified:   src/os_handlers.eliomi
- if the user is already logged in, 
  - if the key is valid, log them in even if it means changing the logged-in account (unchanged)
  - else, ignore the activation key **(very slightly changed)**
- else
  - if the key is valid, log them in (unchanged)
  - else, raise an exception so that a handler will handle that case **(changed)**

# now

Basically, in the action key handling process, in case it's not obvious, an exception is raised and it's up to the user to treat that exception. The template implements a basic version of the handling part.


src/os_handlers.eliom
```diff
@@ -343,6 +355,9 @@ let%shared action_link_handler _myid_o akey () =
     Eliom_registration.(appl_self_redirect Action.send) ()
   | `Custom_action_link (action_link, phantom_user) ->
     Lwt.fail (Custom_action_link (action_link, phantom_user))
+  | `Account_already_activated_unconnected (action_link) ->
+    Lwt.fail (Account_already_activated_unconnected (action_link))
```
vasilisp: Who is going to handle this exception?

pwbs: It should be the user...
In BS, we're going to handle it on our side.
In the template, there should be something to handle it, so we should add something to it.

pwbs:
I forgot to edit os_handlers.eliomi ... :(

pwbs:
I've modified the template accordingly, I think...

pwbs:
So, now, all invalid action keys should be handled by OS users. (The template has been updated accordingly.)

vasilisp:
@pwbs I think we can merge this. Could you rebase?

dannywillems:
I hoped #264 be merged before this PR so it's not me who have to rebase :D.

vasilisp:
If there are no major algorithmic changes, I build up the confidence required to push the green button more easily. So you @dannywillems got lucky. Sorry @pwbs!

pwbs:
ok, this one was accidentally (and automatically) closed. Now https://github.com/ocsigen/ocsigen-start/pull/269/files replaces this one.

PR #258, opened 2016-10-27
Remove useless icons on Android

dannywillems:
Related to https://github.com/ocsigen/ocsigen-start/issues/257.
iOS not done because I can test for the moment.

### DIFF
--- a/template.distillery/Makefile.mobile
+++ b/template.distillery/Makefile.mobile
@@ -325,7 +325,16 @@ $(RUN_PLATFORMS): run-%: check-app-env icons-% assets %
 ##----------------------------------------------------------------------
 ## Rules for the icons and splashes generation.
 
-ANDROID_ICON_NAMES = ldpi mdpi hdpi xhdpi xxhdpi xxxhdpi
+## Android icons names. It corresponds to ldpi, mdpi, hdpi, xhdpi, xxhdpi and
+## xxxhdpi.
+ANDROID_ICON_NAMES = \
+	icon-36 \
+	icon-48 \
+	icon-72 \
+	icon-96 \
+	icon-144 \
+	icon-192
+
 ANDROID_ICONS_PATH = $(MOBILESTATICPATH)/res/android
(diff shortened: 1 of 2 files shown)


template.distillery/Makefile.mobile
```diff
@@ -379,13 +379,29 @@ $(IOS_ICONS_PATH)/icon.png: $(IOS_ICONS_PATH)/icon-57.png; cp $< $@
...
-# Android icons
-$(ANDROID_ICONS_PATH)/ldpi.png: $(ANDROID_ICONS_PATH)/icon-36.png; cp $< $@
-$(ANDROID_ICONS_PATH)/mdpi.png: $(ANDROID_ICONS_PATH)/icon-48.png; cp $< $@
-$(ANDROID_ICONS_PATH)/hdpi.png: $(ANDROID_ICONS_PATH)/icon-72.png; cp $< $@
-$(ANDROID_ICONS_PATH)/xhdpi.png: $(ANDROID_ICONS_PATH)/icon-96.png; cp $< $@
-$(ANDROID_ICONS_PATH)/xxhdpi.png: $(ANDROID_ICONS_PATH)/icon-144.png; cp $< $@
-$(ANDROID_ICONS_PATH)/xxxhdpi.png: $(ANDROID_ICONS_PATH)/icon-192.png; cp $< $@
+$(ANDROID_ICONS_PATH)/ldpi.png: $(ANDROID_ICONS_PATH)/icon-36.png
+	cp $< $@
+	$(RM) $<
```
vasilisp: That seems weird. Next time we need `ldpi.png`, its dependency `$(ANDROID_ICONS_PATH)/icon-36.png` will be missing, so `make` will re-generate something we previously deleted (and we will repeat the same work).

dannywillems: Yes. You're right. Instead, I think I will change the icon names in `config.xml`.

dannywillems:
So I changed. I removed ldpi, mdpi, hdpi, ... and use directly icon-36, icon-48, icon-72, etc in config.xml.

vasilisp:
Thanks! Much nicer this way.

PR #259, opened 2016-10-27
Move all Eliom references in Os_references for consistency.

### LINKED ISSUE #235: Move references in lib in only one file.
There are some references used in the library like `wrong_pdata` and `action_link_key_created` which is in the module `Os_msg`, some others in `Os_userbox`.

There are also references which I don't think they are useful (for example in `Os_userbox`).

I suggest to keep all references (maybe someone can explain why it's useful?) and move all of them in a module `Os_references`.

dannywillems:
Fix #235.

I had no answer so I propose the PR.

### DIFF
--- a/src/os_handlers.eliom
+++ b/src/os_handlers.eliom
@@ -35,7 +35,7 @@ let%server set_personal_data_handler myid ()
     (((firstname, lastname), (pwd, pwd2)) as pd) =
   if firstname = "" || lastname = "" || pwd <> pwd2
   then
-    (Eliom_reference.Volatile.set Os_msg.wrong_pdata (Some pd);
+    (Eliom_reference.Volatile.set Os_references.wrong_pdata (Some pd);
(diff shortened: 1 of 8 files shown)


vasilisp:
In principle I don't think it is a good idea to group objects by their "type". The module structure should follow the "themes" of functionality implemented. Implementation details like references should be internal to the respective modules (or under a `(**/**)` when we absolutely have to break module boundaries).

dannywillems:
@vasilisp I entirely agree with you about objects must be grouped by "themes". But I don't think it's the case for the moment. Maybe for references in `Os_userbox` but not about references in `Os_msg`.

balat:
We don't want to export a module called [Os_references] (provided it implements some kind of references, which is not the case). If you really need to put that outside [Os_userbox] (why?), you should probably give a name related to users or sessions.

PR #262, opened 2016-10-28
Resolve issue about sed -i

### LINKED ISSUE #219: BSD: sed -i error.
In `Makefile.mobile`, we use the command:

```
...
sed -i '1s/^/@charset "UTF-8";/' $(CSS_PREFIX)_$$HASH.css && \
...
```

which causes an error due to `-i` option because `-i` needs an extension.

dannywillems:
Fix #219

### DIFF
--- a/template.distillery/Makefile.style
+++ b/template.distillery/Makefile.style
@@ -1,3 +1,17 @@
+# The following condition is used to determinate if the sed version is BSD or
+# GNU. It is useful because `sed -i` has a different implementation (and this
+# command is needed by $(CSS_DEST). It uses the fact that `sed --version`
+# doesn't exist for the BSD version.
+
+check_sed:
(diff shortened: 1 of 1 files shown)


template.distillery/Makefile.style
```diff
@@ -1,3 +1,15 @@
+# The following condition is used to determinate if the sed version is BSD or
+# GNU. It is useful because `sed -i` has a different implementation (and this
+# command is needed by $(CSS_DEST). It uses the fact that `sed --version`
+# doesn't exist for the BSD version.
+
+check_sed:
```
pwbs: add `check_sed` in `.PHONY`?

pwbs: Also, perhaps these lines should be in `Makefile.options`...

dannywillems: @pwbs Done.

PR #265, opened 2016-10-31
Convert README to README.md and add instructions about mobile applications.

dannywillems:
See https://github.com/dannywillems/ocsigen-start/tree/readme_mobile_instructions/template.distillery for the result.


pwbs:
It's more "convert README to README.md" than "add syntax highlighting..." ^^

dannywillems:
@pwbs You're right. I changed.

vasilisp:
@dannywillems, multiple copies of the same instructions will cause maintenance overhead. Maybe add a link to the tutorial?

dannywillems:
@vasilisp I let @balat to answer. He asked me to do it. He thinks the tutorial is not the right place to put these instructions.

vasilisp:
Fine, but in that case let's remove the tutorial. (Without these instructions, there is nothing left.)

balat:
I agree that it should not be at several places.
But the explanation cannot be ONLY in the tutorial. The tutorial is just a course (that could be done by external party) and do not substitute to manual.
As long as we don't have a manual, it is difficult to know what to put in the tutorial :)
The manual is supposed to be exhaustive but it is probably not the place to put documentation on external project (like android here). May be: explain quickly in manual "install android tools (which ones, where to find documentation)" and give the detailed command in tuto (?)
Let's discuss this during the next Ocsigen meeting :)

PR #266, opened 2016-10-31
Rule in Makefile lib to generate CSS from SCSS. 

dannywillems:
In this [PR](https://github.com/ocsigen/ocsigen-start/pull/147), I mentioned:

> - When uploading the style, the CSS files `static/css/os.css` and `static/css/os_otdemo.css` **must be generated** from `sass/os.scss` and `sass/os_otdemo.scss`. In this way, the CSS is always up-to-date with the equivalent SCSS.

But I didn't provide a rule in the makefile of the library to generate them.

No `--style` option is given to `sass` to have readable CSS files (in the case of the user wants to modify these files and use CSS).

I don't know how the style of the template will evolve but I'm sure some additional SCSS files will have to be added in the `sass` command.


dannywillems:
The second commit update the CSS files `os.css` and `os_otdemo.css` generated by `makefile generate_css`. I checked if the main page wasn't modified and it seems it's OK. I didn't check for the rest.

PR #269, opened 2016-10-31
change behavior for already used activation keys

pwbs:
replacement for #253


pwbs:
don't merge yet... o_O

pwbs:
Ok it's ready now.

PR #270, opened 2016-10-31
OAuth2/OpenID Connect implementation

dannywillems:
**Don't review now. Lots of things must be done.**

Sorry for the reviewer(s) but it's not something we can do in multiple PR. I will provide test applications.


Drup:
If you are going to implement oauth2/openid ... please make an external library. This will get more traction, more visibility and a better library. Several people would be interested by such library.

dannywillems:
@Drup It depends a few on the users table and I created some tables in OS.
I thought about making an external library, but some pieces must remain in OS.

Drup:
> I thought about making an external library, but some pieces must remain in OS.

Of course, there are OS-specific parts that interact with the user tables, but do they have to do with oauth2 really, or just with authentication in general ?

dannywillems:
@Drup I can extract some part which is not OS-specific. Requests, Tokens representation, etc can be in an external library.

dannywillems:
As suggested, I will (later, I don't know when) do an external service for OAuth2.0 and OpenID Connect.

PR #271, opened 2016-11-01
stop using an absolute path for sendmail.

FlorentBecker:
Netsendmail knows about $PATH


vasilisp:
Actually, I think the most correct would be `/usr/lib/sendmail`. It may look strange, but this has been the de facto standard for decades (and `/usr/lib` is not typically in `$PATH`).

PR #286, opened 2016-11-04
Update gitignore template

dannywillems:
I don't know if `.ocp-indent` must be ignored or not.


vasilisp:
> I don't know if .ocp-indent must be ignored or not.

No, I think it should live in the repository to ensure proper indentation by all contributors.

PR #291, opened 2016-11-07
Makefile: use "which" instead of "command" to detect sass (was not wor…

balat:
…king)


vasilisp:
Weird that it is not working for you. `command` is a shell built-in available in at least `dash`, `bash`, and `mksh`. But no big deal, and `which` is also quite standard, so let's merge this.

PR #293, opened 2016-11-07
No error if cordova/res exists

### LINKED ISSUE #292: make android works just the first time
The second time it says:
```
mkdir cordova/res
mkdir: impossible de créer le répertoire «cordova/res»: Le fichier existe
Makefile.mobile:424 : la recette pour la cible « cordova/res » a échouée
```

dannywillems:
Should fix #292


balat:
Danny you're not on holidays! ;) \o/

PR #299, opened 2016-11-09
Add .eliomverbatim file to template

vasilisp:
Nothing but renaming. See ocsigen/eliom#418.

CC @balat


dannywillems:
Why use a suffix and not an hidden file `.eliomverbatim`? @vasilisp

vasilisp:
@dannywillems I had no strong preference between the two, so I just followed @balat's original suggestion. But I have since changed my mind, a file looks cleaner :). Decisions are hard. I will update tomorrow.

PR #304, opened 2016-11-14
remove possibly 1 useless function

pwbs:
`PROJECT_NAME_userbox.connect_form` slightly differs from `Os_view.connect_form`, but I suppose it's just a code-duplication artefact.

### DIFF
--- a/template.distillery/PROJECT_NAME_userbox.eliom
+++ b/template.distillery/PROJECT_NAME_userbox.eliom
@@ -6,31 +6,6 @@
    open Eliom_content.Html.F
 ]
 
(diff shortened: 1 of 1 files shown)


vasilisp:
`Os_view.connect_form` passes `~xhr:false`, in contrast to the one you are removing. I don't remember why, but apparently we have been fine without `~xhr:false` for some time...

pwbs:
Yeah this one is very confusing, that's why I made a separate PR.

Philippe Wang
(mail sent from a phone)

dannywillems:
It was removed in a previous PR. I close.

PR #310, opened 2016-11-23
A few demos in the template

vasilisp:
They need cleanup. Especially the `Os_notif` one; I have not yet figured out a proper style for calling listen in a shared section. Any ideas for additional demos?


balat:
Thanks I'm trying this.
Do we have a demo of Ot_spinner?

vasilisp:
We don't seem to have one.

PR #315, opened 2016-11-24
Demo (also merge OT)

balat:


vasilisp:
Travis takes an eternity, and it is useless in this case. I won't wait!

dannywillems:
@vasilisp It is strange that Travis takes an eternity to begin. I don't know how we can improve it.

PR #321, opened 2016-11-24
Remove error when no sassc

dannywillems:
By the way, `sassc` is not installed on my computer and the error is not working.


vasilisp:
But you do have `sass`, right? From what I understand, it is supposed to fail if neither of them is installed.

dannywillems:
Yes I have `sass`. It's normal the build doesn't fail (because there is no else condition if `sassc` is not installed). This condition is useless if no warning or error is caught.

PR #323, opened 2016-11-24
OPAM depexts: ruby-sass (Ubuntu/Debian), PostgreSQL (Ubuntu/Debien/Mac OSX)

dannywillems:
For mac, `sass` can be installed with `sudo gem install sass` (ruby is already installed with gem). I don't know if we can install it with Homebrew.

About `sassc`, it's not mandatory as `sass` is used if `sassc` is not present.

### DIFF
--- a/opam/opam
+++ b/opam/opam
@@ -20,6 +20,8 @@ depends: [
 ]
 depexts: [
   [["debian"] ["imagemagick"]]
+  [["debian"] ["ruby-sass"]]
   [["ubuntu"] ["imagemagick"]]
+  [["ubuntu"] ["ruby-sass"]]
 ]
 available: [ ocaml-version >= "4.02" ]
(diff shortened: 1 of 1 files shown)


vasilisp:
@balat @dannywillems, Travis takes a loooooong time. I think it is partially because we are bombing them with useless CI jobs and thus create a queue. When modifying only the doc or the template (which do not get CI-checked anyway), please consider adding a `[ci skip]` somewhere in the commit message.

dannywillems:
@vasilisp depext for postgresql is not working in travis because PostgreSQL 9.1 seems to be already installed.

dannywillems:
See https://travis-ci.org/ocsigen/ocsigen-start/jobs/178699644

vasilisp:
APT can be configured to not start daemons on install. See https://askubuntu.com/questions/74061/install-packages-without-starting-background-processes-and-services/501622#501622. We could try to add another Travis `script:` directive for doing this.

dannywillems:
I can resolve it tomorrow.

dannywillems:
Mac OSX builds are also longer than Linux builds.

See #328 for PostgreSQL fix.

vasilisp:
@dannywillems, #328 just merged, thank you!

PR #327, opened 2016-11-24
Add a spinner demo to the template

vasilisp:
It is very simple. I think I have issues with the style (I don't see a spinner, I just see a flashing scrollbar as a side-effect of something). But the DOM logic works.


template.distillery/demo_spinner.eliom
```diff
@@ -0,0 +1,47 @@
...
+(* Class for the page containing this demo (for internal use) *)
+let%shared page_class = "os-page-demo-spinner"
+
+(* Build the spinner *)
+let%client make_spinner () =
+  (* [Ot_spinner.with_spinner_no_lwt] accepts an Lwt thread "slowly"
+     producing HTML content *)
+  Ot_spinner.with_spinner_no_lwt
+    (* sleep for 5 seconds to simulate a delay, then return content *)
+    (let%lwt () = Lwt_js.sleep 500. in
```
vasilisp: Right, thanks :). I used a long delay for debugging and forgot to change it back.

PR #328, opened 2016-11-25
Fix PostgreSQL Travis

dannywillems:
Must work: https://travis-ci.org/dannywillems/ocsigen-start/builds/178786015

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: c
 sudo: required
+before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: c
 sudo: required
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo /etc/init.d/postgresql stop; fi
```
vasilisp: It is generally a better idea to run `service postgresql stop` instead of the script directly (see the `service` manpage). But OK for now :).

Issue #324, opened 2016-11-24
Reference manual


balat:
I think the current version is almost enough.

But:

- [ ] There is an unfinished sentence in Databse section. @vasilisp what did you want to say?
- [ ] We need a short description of the features of the library, that links to the corresponding API doc.

dannywillems:
@balat Maybe use this issue #230?

vasilisp:
For the unfinished sentence, I don't remember what I wanted to say. Maybe I wanted to say that the database can be extended with new tables for the users' needs? I will look into it tomorrow.

PR #329, opened 2016-11-25
Add library overview to the manual

vasilisp:
Also fix DB section. See #324.

`Os_notif` (and possibly others) needs to be documented in detail, I think. `Os_lib` doesn't contain much, and I think it would be a good idea to remove it.

Is there anything we can completely hide from the users?

CC @balat @dannywillems


dannywillems:
@vasilisp I don't agree about removing Os_lib. These module contains functions we can't put elsewhere. And reload function is used in bs I think.

PR #332, opened 2016-11-29
Multi mail

### LINKED ISSUE #317: Fix settings page
- It is not implemented in shared section (thus, broken when generated client-side).
- The style of multi-email is missing

dannywillems:
Fix #317.


dannywillems:
Will be finished for this afternoon or tonight @balat. I reimplement it like on GitHub (I removed the old version I did --').

dannywillems:
New style for emails:
selection_001

dannywillems:
There is also another issue:
- when we add two new emails, the activation link for the second is for the first email. @pwbs Any idea?

vasilisp:
In the future, would it be possible to vertically align the different objects (i.e., one column for mails, another for the status tags, another for the trash bin button)?

But it is not critical. For now, let's just do the minimum amount of work on OS that will unblock the Eliom release.

dannywillems:
@vasilisp Yes it would be better. I'm just focusing on errors for the release.

dannywillems:
@vasilisp Actually, on GitHub, the trash bin and "Set as main email" buttons are on the right. Labels are not aligned.

balat:
I like the still but indeed it should not be centered.
Aligned to the left would probably be better and enough for now?

dannywillems:
I don't like if everything is on the left. You can try by changing `justify-content`.

PR #333, opened 2016-11-29
Os view userbox template

### LINKED ISSUE #318: Improve Os_view
- Rename it to reflect its contents
- Isn't possible to put all function from PROJECT_NAME_userbox.eliom into this module?

dannywillems:
Fix #318.
I merged `Os_userbox`, `Os_view` and `PROJECT_NAME_userbox` in `Os_view`.

About references in `Os_userbox`, I moved them in `Os_user` even if I think it's not the best solution (I prefer `Os_references` as I did before, or at least put all references in one module).


dannywillems:
I moved everything in `Os_view`. I know @balat doesn't agree with that (`Os_view` is too big) but it's more clearer than before.
`reset_tip_links`, `upload_link_pict` has moved in `Os_view`.

I also add parameters for texts in a function. I think it must be generalized for others.
If it's useful before merging, I will do it in this PR.

balat:
No I'm ok with that.
I just think we should not call it Os_view.
Because we cannot put the "view" of the whole app in one file,
and (if I'm right) this file only concerns user registration and login.

balat:
@dannywillems I'd like to merge this one. Can you fix the last problems asap?

Issue #326, opened 2016-11-24
Change avatar in settings.


dannywillems:
- The avatar is not resized.
- Style not done.

dannywillems:
@balat I think we can remove the binding to imagemagick at the same time (see #231) with the solution you use in bs.

balat:
@sagotch ?

dannywillems:
@sagotch realized it's just a CSS issue for resizing. About cropping the image server-side, we can use:
```
Lwt_process.exec
  ("", [|"convert"; "+repage" ;
         "-strip"; "-interlace"; "Plane"; "-auto-orient";
         "-define"; Printf.sprintf "jpeg:size=%dx%d" (2 * w) (2 * h);
         "-resize"; Printf.sprintf "%dx%d!" w h; "-quality"; "85";
         "-background"; "white"; "-flatten";
            (* In case of transparent image *)
         input_file;
         "jpg:" ^ output_file|])
```
as suggested in https://github.com/ocsigen/ocsigen-start/pull/335.

Issue #227, opened 2016-10-18
[Template] Style


dannywillems:
Do the style of the entire template.

balat:
Almost finished. multi-email still to be fixed.

PR #335, opened 2016-11-30
Remove ImageMagick binding dependency

### LINKED ISSUE #231: Remove dependency to the binding to ImageMagick.
@vouillon suggests to remove the dependency to the binding to ImageMagick which is only used in one function (`Os_uploader.crop_image`). Instead of using the binding, the binaries can be called directly.

dannywillems:
Fix #231.

`Os_uploader` was never used...

@balat @vouillon Cropping is totally broken and Ocsigen Start uses `Ot_picture_uploader` which is not simple (and I think it's done client-side? Possible?). Work must be done for that! Please, take a look.

There are related issues (like #326), and first version can not be released without this! We can not finish the settings page (#317) and the style (#227, related to settings page). I remember you have an elegant and working solution with `convert` and by using `Sys.command`.

A simple `convert` use can be `convert [image] -crop 120x120+50+50 [out_image]` to have a 120x120 square and it must be done server-side (I don't know how it's done in `Ot_picture_uploader`).

I can propose a simple solution but I think it's useless if you have already the appropriate solution.


vasilisp:
@balat @vouillon can you share your `convert` magic?

vouillon:
We are using this:
```ocaml
Lwt_process.exec
  ("", [|"convert"; "+repage" ;
         "-strip"; "-interlace"; "Plane"; "-auto-orient";
         "-define"; Printf.sprintf "jpeg:size=%dx%d" (2 * w) (2 * h);
         "-resize"; Printf.sprintf "%dx%d!" w h; "-quality"; "85";
         "-background"; "white"; "-flatten";
            (* In case of transparent image *)
         input_file;
         "jpg:" ^ output_file|])
```
and check that the process returns `Unix.WEXITED 0`.

PR #338, opened 2016-12-07
Client-side message for failed connection

vasilisp:
While trying to sign up, in case `sign_up_handler` fails (e.g., because the DB is not running), display a pop-up.

For now, the previous pop-up stays. To close it, the handler would need access to the corresponding `close` function, and obtaining that seems to be quite messy.

@balat will this do for now?

### DIFF
--- a/src/os_handlers.eliom
+++ b/src/os_handlers.eliom
@@ -156,7 +156,25 @@ let%client sign_up_handler_rpc =
        sign_up_handler_rpc)
 
 let%client sign_up_handler () v =
-  sign_up_handler_rpc v
+  let error msg =
+    let%lwt _ =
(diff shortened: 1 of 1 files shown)


balat:
I guess it's enough. This should not happen too often.
And it's ok if the sign up popup is not closed.

But there is a problem with the stylesheet.

dannywillems:
I'm not sure it must be added in the library.

vasilisp:
@dannywillems, I think you are right. It is just a temporary hack for a prominent issue that users of 0.99 may be confronted with. We do need to re-think DB error management.

vasilisp:
Following discussion elsewhere, I am closing this. We will think about a better and more general solution.

Issue #334, opened 2016-11-30
Demo: example with DB


balat:
An example of page generated from db, using PGOCaml. For example display the list of users.
Also fix makefiles.

dannywillems:
What do you mean by fixing makefiles? @balat

balat:
I mean that it does not compile :)
The solution is probably to do
```
export DB_HOST
export DB_PORT
etc.
```
and remove ```ENV_PSQL```
(at least we won't see the password in the terminal).

Issue #325, opened 2016-11-24
When no database, no message about it.


dannywillems:
When the database has not been created, it's not possible to sign up. When we click on the `send` button in the sign up popup, nothing happens and there is no message in the log.

dannywillems:
This issue was resolved by the pgocaml demo.

PR #339, opened 2016-12-08
Demo (merge OT too)

balat:


dannywillems:
It will fix #334 and #325 (with the PR #338).

PR #341, opened 2016-12-08
Fix avatar resizing

### LINKED ISSUE #326: Change avatar in settings.

dannywillems:
Depends on https://github.com/ocsigen/ocsigen-toolkit/pull/104 and https://github.com/ocsigen/ocsigen-toolkit/pull/103.
Fix #326.

~~#335 can also be merged because we don't use it anymore. Cropping is done client-side.~~

~~Cropping server-side needs the command given in #335. It is useful if we want to save the initial image server-side for example in the database or for future cropping of the same image. Not useful for first release.~~

### DIFF
--- a/template.distillery/sass!os.scss
+++ b/template.distillery/sass!os.scss
@@ -277,3 +277,12 @@ a.os-page-header-app-name {
         }
     }
 }
+
+.os-avatar {
+    margin-right: 15px;
+    width: 35px;
+}
+
+.ot-dr-left .os-avatar {
+    margin-left: 16px;
+}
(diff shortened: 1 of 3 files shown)


vasilisp:
@dannywillems I haven't run it to see how the style changes, but the code looks reasonable. Can you rebase?

dannywillems:
@vasilisp I will wait for #346 because it would have new conflicts I think.

dannywillems:
Done. Can be merged.

template.distillery/sass!os.scss
```diff
@@ -276,4 +276,13 @@ a.os-page-header-app-name {
...
     }
+}
+
+.os-avatar {
+    margin-right: 15px;
+    width: 35px;
+}
+
+.ot-dr-left .os-avatar {
+    margin-left: 16px;
```
dannywillems: This line is here to have the same space on the left than the elements in the drawer.

PR #352, opened 2016-12-09
Fix match failure

balat:


balat:
I merge this myself, as it it completely broken without.

PR #355, opened 2016-12-09
Remove ImageMagick binding dependency and use convert command directly

### LINKED ISSUE #231: Remove dependency to the binding to ImageMagick.

dannywillems:
Fix #231

### DIFF
--- a/src/os_uploader.eliom
+++ b/src/os_uploader.eliom
@@ -18,22 +18,84 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
 
+exception Error_while_cropping of Unix.process_status
+exception Error_while_resizing of Unix.process_status
+
+let%server resize_image ~src ?(dst = src) ~width ~height =
+  let%lwt resize_unix_result =
+    Lwt_process.exec
+      ("", [|"convert";
+             "+repage" ;
+             "-strip";
+             "-interlace"; "Plane";
+             "-auto-orient";
+             "-define"; Printf.sprintf "jpeg:size=%dx%d" (2 * width) (2 * height);
+             "-resize"; Printf.sprintf "%dx%d!" width height; "-quality"; "85";
+             (* In case of transparent image *)
+             "-background"; "white";
+             "-flatten";
+             src;
+             "jpg:" ^ dst
+           |]
+      )
+  in
+  match resize_unix_result with
+  | Unix.WEXITED status_code when status_code = 0 -> Lwt.return ()
+  | unix_process_status -> Lwt.fail (Error_while_resizing unix_process_status)
+
+let%server get_image_width file =
+  let%lwt width =
+    Lwt_process.pread
+      ("", [| "convert" ;
+              file ;
+              "-print" ; "%w" ;
+              "/dev/null"
+           |]
+      )
+  in
+  Lwt.return (int_of_string width)
+
+let%server get_image_height file =
+  let%lwt height =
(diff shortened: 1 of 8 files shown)


vasilisp:
@dannywillems thanks! I haven't read the code carefully yet. I will look at it tomorrow or on Monday (latest).

src/os_user_view.eliom
```diff
@@ -259,14 +258,16 @@ let%shared upload_pic_link
...
+            ~after_submit:close upload_service) ;
       Lwt.return ()
     with e ->
       Os_msg.msg ~level:`Err "Error while uploading the picture";
       Eliom_lib.debug_exn "%s" e "→ ";
       Lwt.return () ) : _ ) ] :: a) content
 
-let%shared reset_tips_link (close : (unit -> unit) Eliom_client_value.t) =
+let%shared reset_tips_link
+    ?(close : (unit -> unit) Eliom_client_value.t = [%client (fun () -> () : unit -> unit)]) ()
```
dannywillems: Same comment than `upload_pic_link`.

template.distillery/PROJECT_NAME_settings.eliom
```diff
@@ -108,10 +107,10 @@ let%shared settings_content () =
         Os_user_view.password_form ~service:Os_services.set_password_service ();
         br ();
         Os_user_view.upload_pic_link
-          none
+          ~submit:([a_class ["button"]], [pcdata "Submit"])
```
dannywillems: Just for style.

vasilisp:
Shouldn't you downsize the avatar? If I choose a big area, it is displayed in 100% magnification and the layout looks very broken.

dannywillems:
@vasilisp It must be fixed by a previous by #341. Or you mean the cropping section?

dannywillems:
@vasilisp See #363.

src/os_uploader.eliom
```diff
@@ -18,22 +18,86 @@
...
+  let%lwt width_src = get_image_width src in
+  let%lwt height_src = get_image_height src in
+  let left_px = pixel_of_percent left width_src in
+  let top_px = pixel_of_percent top height_src in
+  let width_cropped = width_src - left_px - (pixel_of_percent right width_src) in
+  let height_cropped = match ratio with
+    | None -> height_src - top_px - (pixel_of_percent bottom height_src)
+    | Some ratio -> truncate ((float_of_int width_cropped) /. ratio)
+  in
+  let%lwt crop_unix_result =
```
vasilisp: These conversions worry me a bit. I think it would be cleaner to do the `right - left` subtraction (same for the vertical coefficients) and then the convertion to pixels, saving a few truncations.

But it was similar before, so we could merge it like this.

src/os_uploader.eliom
```diff
@@ -18,22 +18,86 @@
...
+    Lwt_process.exec
+      ("", [|"convert";
+             "-crop" ;
+             Printf.sprintf "%dx%d+%d+%d" width_cropped height_cropped left_px top_px;
+             src ;
+             dst
+           |]
+      )
```
vasilisp: Have you tested for graceful failure if `convert` et al. are not there? That would be nice :).

dannywillems: @vasilisp No I have not tested.

vasilisp:
We can merge this and improve it later :).

PR #356, opened 2016-12-09
Install NPM dependencies [skip ci]

### LINKED ISSUE #302: install npm packages as part of build
I keep hitting errors due to missing npm packages. First issue was that `postcss` wasn't installed. I installed that, then updated the Makefile as needed. Then I got errors about `sassc` and `autoprefixer` not being found.

You might want to consider the Makefile I used for [cufp.org](https://github.com/CUFP/cufp.org). The main points are:
- Assume all NPM packages will be installed locally within the repo, thus eliminating any assumptions about the users global NPM state.
- Define paths to all executables up at top, e.g. `postcss=node_modules/.bin/postcss`.
- Define rules to install 3rd party packages. See the section so titled.
- Add NPM packages as pre-requisites to rules needing them, e.g. see the rule for generating `site/css/app.css`.

The net result is that a simple `make` does a lot more to help people get started.

dannywillems:
Fix #302.

It's not the best solution but I think it's enough for the moment.


vasilisp:
This way, the template becomes completely unusable without NPM.

dannywillems:
@vasilisp Yes, you're right... I don't have any good (and elegant) idea to implement it.

balat:
This is urgent. What do we do? The current solution (without this PR) is not acceptable.

balat:
I merged this. Feel free to propose a solution that works without npm ...

PR #361, opened 2016-12-12
i18n integration (second test).

### LINKED ISSUE #84: Add i18n
Using module https://github.com/besport/ocsigen-i18n
For example, translate all error messages ("wrong password", etc) in french.
Also detect the language from browser settings or mobile device (ask Besport about that).

dannywillems:
Fix #84.

## Dependency.
- [x] ocsigen-i18n: https://github.com/besport/ocsigen-i18n/pull/17 must be merged first.
- [x] add ocsigen-i18n in OPAM, README, Jenkins, Travis.

## Manual.
- [x] Write a manual.

## Template:
- [x] Add in makefiles the generation of i18n file.
- [x] Add a tsv sample file. Added in assets directory in the template. **Text must be added. See `templates.distillery/assets!PROJECT_NAME_i18n.tsv`.**
- [x] Use the PPX extension for the translation.
- [x] Add a language template file to get a sum type with languages (see https://github.com/besport/ocsigen-i18n) --> `PROJECT_NAME_language.eliom[i]`.
- [x] Update SQL file by adding a field `language` in `ocsigen_start.users` table.
- [x] Documentation, documentation and documentation!!!
- [x] Test, test and test!!!
- [ ] Add a button to change the language.

## Library:
- [x] Add a field language in the `ocsigen_start.users` tables.
- [x] Update functions which access to users table.
- [x] Update type `Os_types.users`.
- [x] Detect the user language when starting a process. If not connected, based on Accept-Language value in the HTTP header. If no value is given in the header, the default value is `"en-US"` and it can be changed with `Os_language.set_default_language`. If the user is connected, it retrieves the language from the database.
- [x] `Os_userbox` and `Os_handlers` depend on some translations, which are not available in the library (because only available in the template. Even if it's ugly, `Os_i18n` contains references to the text and they are overwritten by the template for the appropriate translations (see `PROJECT_NAME_language`).
## Arbitrary choice
- The language can be NULL in the table. It implies:
  - When creating/updating an user, the language is optional. See `Os_db.User.update` and `Os_db.User.create`. 
- `set_personal_data` doesn't require the language. The language is not considered as a personal data.
## Issue.
- [x] `eliomdep` is executed before `make i18n-update`

# Not done.
- internationalize texts in handlers. It's not easy due to RPC. Text used with `Os_msg` is still in English.
- Avoid calling `make i18n-update` manually: need to be called when compiling. It implies some issues with `eliomdep` calls.


dannywillems:
It's working now!
Need to add a button to change the language in settings.

balat:
I can't buid a freshly generated project.
It seems that depedencies are not generated correctly because `PROJECT_NAME_i18n.eliom` is missing.
Then `PROJECT_NAME_i18n.eliom` uses and undefined variant `Us` ...
Do you have the same problems?

dannywillems:
@balat You need to generate the file `PROJECT_NAME_i18n.eliom` by using `make i18n-update`.

balat:
Yes I did it. But it would be better if it works without one more step.

And I still have the second problem:
```
File "bone_i18n.eliom", line 4, characters 42-44:
Error: Unbound constructor Us
```

dannywillems:
For the first error, it must be due to `ocsigen-i18n`: you need to update it (you can unpin it as it's in OPAM).

For the second, it's fixed.

dannywillems:
Waiting for https://github.com/besport/ocsigen-i18n/pull/20 and https://github.com/besport/ocsigen-i18n/pull/22.

vasilisp:
I think we can get rid of the manual generation of the `_i18n.eliom` file, see https://github.com/ocsigen/ocsigen-start/commit/b06b71d6073b4184f3e3b020799929e9dcb80015 .

dannywillems:
@vasilisp The problem I had it's about executing `eliomdep` before compiling.
Else, it's a good solution. I also suggest to have a rule `i18n-update` (with `$(I18N_ELIOM_FILE)` as dependency) to have a target easier to remember.

sagotch:
i18n: `you_click_on` is wrong.

vasilisp:
@dannywillems, the rules involving `eliomdep` depend on `$(SERVER_FILES)`, which includes `$(I18N_ELIOM_FILE)`. The latter will thus be generated automatically by my modified rule.

dannywillems:
I close since #399.

PR #380, opened 2016-12-22
add geolocation plugin

pwbs:
@balat @dannywillems

### DIFF
--- a/template.distillery/mobile!config.xml.in
+++ b/template.distillery/mobile!config.xml.in
@@ -101,10 +101,10 @@
     <preference name="KeyboardShrinksView" value="false" />
 
     <!-- For geolocation: -->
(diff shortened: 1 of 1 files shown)


dannywillems:
Maybe give a link to the binding?

pwbs:
> Maybe give a link to the binding?

I don't know... In my experience, the plugin is broken (at least for iOS), hence the binding is useless (for now).
That being said, it doesn't cost much to give a link to the binding...

pwbs:
> > Maybe give a link to the binding?

> I don't know... In my experience, the plugin is broken (at least for iOS), hence the binding is useless (for now).
That being said, it doesn't cost much to give a link to the binding...

Done...

PR #400, opened 2017-01-12
App under path

vasilisp:
A set of hacks that enable us to put the application under a path.

The path needs to be provided via the command line:

```
gmake emulate-android APP_REMOTE=yes APP_SERVER=http://192.168.1.1:8080 APP_PATH=/foo
```

The `APP_PATH` must be provided with the leading slash but with no trailing slash. `APP_PATH` is entirely optional. The `.conf.in` needs to be modified manually (`<site>` tag).

There is a corresponding branch for Eliom (`app-path`).

### DIFF
--- a/src/os_handlers.eliom
+++ b/src/os_handlers.eliom
@@ -188,9 +188,7 @@ let%client restart ?url () =
            (Js.string url)
        | None ->
          ());
-     Eliom_client.exit_to ~absolute:false
-       ~service:(Eliom_service.static_dir ())
-       ["eliom.html"] ())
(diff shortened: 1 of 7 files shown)


vasilisp:
There is still a small issue with client routing (e.g., as used to restart the app). Client routing looks at the whole path, without throwing away the prefix that belongs to the app and not to the individual services. Fix coming (in Eliom).

vasilisp:
The client routing and restart trouble is resolved, I think.

balat:
We probably want a small note about this in README?
It is not easy to find, isn't it?

PR #411, opened 2017-01-17
npm install cordova and friends

vasilisp:
We automatically (and locally) install `cordova` and related packages (addendum to #356).

Currently, I have trouble with

- the compilation of CrossWalk
- the compilation process complaining that I haven't accepted Android licenses (similar to [this StackOverflow discussion](https://stackoverflow.com/questions/40383323/cant-accept-license-agreement-android-sdk-platform-24))

The first goes away if I disable CrossWalk, and the second if I manually install the Android packages that `gradle` tries to install.  Neither is related to the PR (or our code) per se, but with the combination of external packages that get installed. But we need to resolve the issue anyway.


vasilisp:
@balat this is ready to merge, I think.

If anybody is willing to test with both `USE_NPM = yes` and `USE_NPM = no` (the latter assuming system-wide-installed packages) in `Makefile.options`, that would be useful.

balat:
Great!
I'm testing this.

PR #422, opened 2017-01-26
Use externally defined settings service in action link handler

vasilisp:
We need this to redirect to the settings page after validating a new e-mail. For this, we need access to the settings service (defined in the template) from the lib. See the comments for details.


vasilisp:
Implementation updated. Now the settings service is used to generate the validation link.

PR #428, opened 2017-02-09
going towards enhancing performance in database connection usage

pwbs:
modified:   src/os_db.ml
	modified:   src/os_db.mli
	modified:   src/os_group.ml
	modified:   src/os_group.mli
	modified:   src/os_session.eliom


vouillon:
This is not just for performance issue. This makes it possible to use these functions while already holding a database handle. Without that, we can get a deadlock when these functions try to acquire another database handle.

pwbs:
There are possibly other functions that require this kind of work.

PR #439, opened 2017-02-21
Fix doc header for os_handlers.

dannywillems:
See http://ocsigen.org/ocsigen-start/dev/api/client/Os_handlers.
I'm not sure it's caused by the missing empty line but I remember I resolved a similar issue like that.


vasilisp:
I don't know if the empty line is needed (add one to make sure), but I am pretty sure that the doc comment has to be after `[%%shared.start]`. See https://github.com/ocsigen/ocsigen-toolkit/issues/114.

dannywillems:
@vasilisp Changed.

PR #445, opened 2017-03-15
compatibility: interface change in Ot_carousel

jrochel:
depends on: https://github.com/ocsigen/ocsigen-toolkit/pull/129


jrochel:
The interface has not changed after all.

PR #454, opened 2017-05-09
README: fixed markdown syntax for titles.

sagotch:


vasilisp:
It now seems to render fine. Thanks!

PR #461, opened 2017-07-12
Modular jsoo : add progressvie web app template

aik9508:
I add a new template (pwa.ppx) in ocsigen-start. It's very similar with `basic.ppx` template in eliom except for the new-added service_worker.ml and some tiny modifications in Makefile.options and Makefile.


jrochel:
Both your commits introduce white spaces at the end of lines. Git's syntax highlighting complains about it, and I do too.

PR #465, opened 2017-08-16
Phone connectivity

vasilisp:
This PR allows users to create an account with their phone number.

The "core" of this functionality lies in the `Os_connect_phone` module, which allows sending verification codes via SMS and permits creating an account once the user validates the code.

Phone connectivity is entirely optional. It can be disabled by removing the `PROJECT_NAME_phone_connect.eliom` file.

The template has been updated to reflect the changes in the UI. UI to be improved later.

CC @balat @vouillon @jrochel @pwbs


src/os_connect_phone.eliom
```diff
@@ -0,0 +1,119 @@
...
+type%shared sms_error_core = [`Unknown | `Send | `Limit]
+type%shared sms_error = [`Duplicate_number | sms_error_core]
+
+(* adapted from Ocsigen_lib.make_cryptographic_safe_string *)
+let activation_code =
+  let rng = Cryptokit.Random.device_rng "/dev/urandom" in
+  fun () ->
+    let random_number = Cryptokit.Random.string rng 20 in
+    let to_b64 = Cryptokit.Base64.encode_compact () in
+    (* CHECKME: is this cryptographically safe? probably not *)
```
vouillon: That's 36 bits, which seems fine if we limit the number of connection attempts.

src/os_lib.eliom
```diff
@@ -102,6 +102,13 @@ module Email_or_phone = struct
         s, `Almost_email
       else if Re_str.string_match phone_regexp s 0 then
         let s = string_filter ((<>) ' ') s in
+        let s =
+          (* Also accept 00 prefix and normalize to + *)
```
vouillon: Note that '00' is not universal.

vasilisp: I could remove this for now, or leave it there as a small optimization for most Europeans (hard to imagine it causing trouble to anyone else). What do you think?

In the future we could get fancier and use geolocalization to guess the user's intention.

src/os_connect_phone.eliom
```diff
@@ -0,0 +1,175 @@
...
+       confirm_code_extra)
+
+let%server confirm_code_signup
+    (firstname, lastname, code, password) =
+  match%lwt Eliom_reference.get activation_code_ref with
+  | Some (number, code', _) when code = code' ->
+    let%lwt () = Eliom_reference.set activation_code_ref None in
+    let%lwt user = Os_user.create ~password ~firstname ~lastname () in
+    let userid = Os_user.userid_of_user user in
+    let%lwt () = Os_db.Phone.add userid number in
```
vouillon: Note that this will fail if the number was taken in the meantime.

src/os_lib.eliom
vouillon: We can keep this for now

PR #468, opened 2017-08-23
Explain a beginner's mistake about "shared signals"

ghuysmans:
Maybe it doesn't belong there, I don't know...

### DIFF
--- a/template.distillery/demo_calendar.eliom
+++ b/template.distillery/demo_calendar.eliom
@@ -21,6 +21,19 @@ let%server service =
 let%client service = ~%service
 
 (* A reactive value containing the currently selected date *)
+(* NOTE: in this example, we define a shared signal on the server side. Its
+   original value can only be read when the server generates the first page
+   (declaring it `%client`-only would obviously not work) and injected to be
+   read-/writable on the (possibly disconnected) client side since any
+   *shared value* is injectable; subsequent updates won't be sent to the server.
+   Declaring this signal as `%shared` wouldn't work either, as you'd end up with
+   two different signals (one for each side): a Reactive `map` in `page` would
+   use the server's signal when it's first generated on the server, while the
+   client-side click event would use its own `f`, so nothing would actually
+   happen. You can observe this duplication by replacing `%server` below with
+   `%shared`: the compiler will emit an error because the type of one of those
+   signals can't be inferred (it remains unknown at the end of the typing pass)
+   since it's never used throughout the program. *)
 let%server s, f = Eliom_shared.React.S.create None
 
 let%client action y m d = ~%f (Some (y, m, d)); Lwt.return_unit


template.distillery/demo_calendar.eliom
```diff
@@ -21,6 +21,16 @@ let%server service =
 let%client service = ~%service
 
 (* A reactive value containing the currently selected date *)
+(* NOTE: in this example, we want the signal to be stored on the server side,
+ * so that any change will be sent to the server. Only declaring this signal as
```
vasilisp: That's not very accurate. The shared signal is defined on the server so that we can use it to produce the first page (and along with it, send a real signal that the client will use to update the page). This has nothing to do with sending changes to the server.

template.distillery/demo_calendar.eliom
```diff
@@ -21,6 +21,16 @@ let%server service =
...
+ * shared won't work since you'll end up with two different signals (one for
+ * each side): if you only use a Reactive `map` in `page`, it will use the
+ * server's signal when it's first generated on the server, while the
+ * client-side click event will use its own `f`, so nothing will actually
+ * happen. You can observe this duplication by replacing the `%server` below
+ * with `%shared` and the error the compiler produces: the signal's type can't
+ * be inferred since once instance is never used throughout this program.
+ *)
```
vasilisp: The rest looks good. Could you just remove the stars in the beginning of each line? We generally avoid them because they make re-indenting the code harder.

ghuysmans:
Maybe the injection below (in `action`) should be documented too since functions normally can't be injected...

vasilisp:
Why not. If you can add a comment stating that it is not a plain server-side function but a shared value containing a function (shared values being injectable), that would be great.

template.distillery/demo_calendar.eliom
ghuysmans: Oh, I see... that's the precise point I hadn't got yet!

ghuysmans:
I think it's way better now...

template.distillery/demo_calendar.eliom
```diff
@@ -21,6 +21,18 @@ let%server service =
 let%client service = ~%service
 
 (* A reactive value containing the currently selected date *)
+(* NOTE: in this example, we define a shared signal on the server side. It can
+   be accessed when the server generates the first page (so, declaring it
```
vasilisp: > It can be accessed when the server generates the first page

Yes and no. It's value can be accessed, but it cannot be used as a true signal (i.e., updated). What about something like:

> The original value of the signal is accessible on the server while generating the first page

vasilisp:
Yes, please use `-f` for minor updates of the preceding commit. A separate commit wouldn't be interesting enough to keep in Git history.

template.distillery/demo_calendar.eliom
ghuysmans: Not updatable since it's wrapped in a `Eliom_shared.Value.t`? Is it worth mentioning it there? Maybe parts of this should be moved to the corresponding documentation page...

vasilisp: Not updatable because the implementation does not use a server-side reactive signal at all (but only the fixed value of the signal). If curious, see [`Eliom_shared`](https://github.com/ocsigen/eliom/blob/master/src/lib/eliom_shared.eliom#L272).

For whatever is missing from [the documentation page](http://ocsigen.org/eliom/6.2/manual/clientserver-react), I will be happy to review PRs :).

PR #469, opened 2017-08-29
Os_email.send: separate url from the rest of the content

vasilisp:
Useful for customizing the sent e-mail.


vasilisp:
CC @vouillon.

PR #471, opened 2017-08-29
Add page transition demo in the template

aik9508:
Add page transition demo in the template


balat:
The compilation fails when we follow the instructions from README.
```
Fatal error: exception Fl_package_base.No_such_package("cordova-plugin-screenshot", "")
```

I think it is the first time the plugin uses an OCaml dependency that is not already a dependency for Ocsigen Start. It must be documented in README.

PR #478, opened 2017-10-18
Os_db.init: no pool parameter means no pool at all

ghuysmans:


vasilisp:
Do you have a use-case for this? Why is no pool preferable to a sane default pool size?

ghuysmans:
I think the current use of `None` is misleading. We've measured that establishing a lot of connections is not a good thing to do, but the interface doesn't tell you about the arbitrary minimum pool size.

PR #488, opened 2017-12-22
OCaml 4.06 compatibility fix (-safe-string)

ghuysmans:


ghuysmans:
There are missing dependencies (for the template)

PR #490, opened 2018-01-08
Remove ocaml-plugin-screenshot dependency

vasilisp:


balat:
To be discussed:
 - either we accept this and we find a way to put it back very soon
 - or we find an alternative for the screenshot feature (like: animation with blank screen instead)
 - or we release properly our plugin binding (10 line intreface written by Be Sport)

PR #502, opened 2018-04-19
Fix link without versions numbers with symlink

CelieArnaud:


vasilisp:
- http://ocsigen.org/ocsigen-start/manual/ works
- http://ocsigen.org/ocsigen-start/1.1.0/manual/ works
- API links like http://ocsigen.org/ocsigen-start/api/server/Os_date work
- http://ocsigen.org/ocsigen-start/api/ doesn't work, but I don't remember if it ever worked. Separate issue anyway.

Thanks!

PR #504, opened 2018-05-03
change reload behavior to use current url

remyzorg:


remyzorg:
It seems my pull request breaks some reloads... So we should not merge this now. I'll work on it further.

balat:
I think this is hazardous as it breaks Eliom's service logic.
I would prefer to understand first in which cases current solution does not work.
We can probably fix them ...

PR #514, opened 2018-07-20
Serve pages as xhtml by default

balat:


vasilisp:
I was away and haven't followed the discussion on this. Is it still relevant, or have we decided against it?

balat:
This makes it possible to have nested <a> tags, which is very useful.
But we decided not to apply this at Be Sport because it breaks tweet inclusion code from twitter ...

PR #516, opened 2018-08-21
update version of resource-pooling dependency

jrochel:
This ensures compatibility with Lwt 4.
I'm not sure whether this is actually necessary. I'll put my trust in the reviewer.


vasilisp:
I don't think we need this. If you have Lwt 4.x, the OPAM solver will automatically find a solution with `resource-pooling >= 0.3.1`; if you have Lwt 3.x, you are free to use older versions of `resource-pooling`.

PR #532, opened 2019-01-07
Fix CI's exit code and add redirection to demo on ocsigen-1.inria.fr.

leovalais:


leovalais:
It is OK to merge because the CI now fails *as expected* as there are incorrect wikis inside `doc`.

PR #533, opened 2019-01-07
fix: add quotes for [[wiki("which")... syntax

leovalais:


leovalais:
@balat you can merge this. The new intro should appear.

PR #564, opened 2019-10-04
Os_date.now: Make it possible to specify the timezone server side

ilankri:


opam
```diff
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "ocsigen-start"
-version: "2.5.0"
+version: "2.5.1"
```
balat: Interface change → 2.6.0

ilankri: Thanks!

PR #572, opened 2019-12-13
Completely switch to PPX

vouillon:


src/os_db.ml
```diff
@@ -1,663 +1,3115 @@
...
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program; if not, write to the Free Software
- * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
- *)
-
+(* GENERATED CODE, DO NOT EDIT! *)
```
balat: Do we want to include it in git?

vouillon: Do you see an alternative?

PR #573, opened 2019-12-20
Add the counterpart function of set_tip_seen

Zzull:


Zzull:
Does this warrant a bump in the version?

PR #601, opened 2020-09-16
Add a raw unconnected phone code confirmation function

remyzorg:


remyzorg:
@vouillon ok it works as expected, thank you again.

PR #607, opened 2020-10-13
[template] split demo i18n entries to separate file

facelesspanda:
:warning: Makes use of unpublished ocsigen-i18n@3.7.0. We should release ocsigen-i18n before merging this. (Or at least before releasing it)

Cleaning out the demo after installing the template is always a pain, because all the i18n entries are in a single file. Now, we can split them up, so it can be less of a pain.

I'm adding a dependency to `ocsigen-i18n`. I'm not sure if that's the right thing to do because ocsigen-start itself doesn't depend on it, it's just the project generated from the template. Maybe we should have an opam file for the template?


jrochel:
Opam PR for ocsigen-i18n: https://github.com/ocaml/opam-repository/pull/17402

PR #609, opened 2020-10-14
[Demo] Another fix in demo_notif

balat:
We must use React.E.map only once if we come to the page several times


balat:
Already merged in #610

PR #612, opened 2020-10-20
[Makefiles] restore cmd-pipe directory creation

balat:


template.distillery/Makefile.os
```diff
@@ -46,7 +46,8 @@ endif
 .PHONY: all css byte opt
 
 DIST_DIRS          := $(ETCDIR) $(DATADIR) $(LIBDIR) $(LOGDIR) \
-                      $(FILESDIR)/avatars/tmp $(ELIOMSTATICDIR)
+                      $(FILESDIR)/avatars/tmp $(ELIOMSTATICDIR) \
+                      $(shell dirname $(CMDPIPE)) $(ELIOMTMPDIR)
```
jrochel, line 50: Where is `ELIOMTMPDIR` defined?

balat: Indeed, ELIOMTMPDIR seems to be set to /tmp in config file without any variable ...
It is probably not needed. I am still not completely sure for static dir but at least it compiles now.

jrochel:
https://github.com/ocaml/opam-repository/pull/17448

PR #616, opened 2021-01-13
Os_tips: Change the type of 'onclose'

Zzull:
I need to animate with `Lwt` the closing of a tip

### DIFF
--- a/opam
+++ b/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "ocsigen-start"
-version: "3.0.0"
+version: "4.0.0"
 authors: "dev@ocsigen.org"
 maintainer: "dev@ocsigen.org"
 synopsis: "An Eliom application skeleton ready to use to build your own application with users, (pre)registration, notifications, etc"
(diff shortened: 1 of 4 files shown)


opam
```diff
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "ocsigen-start"
-version: "3.0.0"
```
jrochel, line 3: Pour un breaking change il suffit d'augmenter la version mineur : 3.1.0

jrochel (changes requested):

opam
jrochel: (Si je ne me trompe pas. C'est peut-être différent pour Ocsigen.)

Zzull: What warrants a major version then?

jrochel: Some sort of paradigm shift for example? Again, I vaguely remember @balat saying that for Ocsigen every breaking change requires an increase of the major version.

vouillon: https://semver.org/
> Given a version number MAJOR.MINOR.PATCH, increment the:
>
>    MAJOR version when you make incompatible API changes,
>    MINOR version when you add functionality in a backwards compatible manner, and
>    PATCH version when you make backwards compatible bug fixes.

PR #624, opened 2021-04-28
[ci] Automate the build process using GitHub Actions instead of Travis CI

ilankri:


smorimoto:
I recommend using v2 instead.

ilankri:
> I recommend using v2 instead.

But this version is not yet released, no? I am a little uncomfortable with using an unreleased version.

smorimoto:
> But this version is not yet released, no? I am a little reluctant to use an unreleased version.

Well, that's true, but v2 actually includes a lot of fixes and improvements.

smorimoto:
Since it was officially released, I changed it to use ocaml/setup-ocaml instead :)

smorimoto (approved):

PR #625, opened 2021-05-03
[tips] prioritized bubble opening

clembu:
Adds a `priority` parameter to `Os_tips.bubble`,
and changes the tips system to open bubbles on a given order.

Behaviour on absent `priority` is unchanged.

### DIFF
--- a/src/os_tips.eliom
+++ b/src/os_tips.eliom
@@ -189,12 +189,101 @@ let%shared block ?(a = []) ?(recipient = `All)
 let%client onload_waiter () =
   let%lwt _  = Eliom_client.lwt_onload () in Lwt.return_unit
 
-(* This thread is used to display only one tip at a time *)
-let%client waiter = ref (onload_waiter ())
+(* This list of threads is used to display only one tip at a time,
+   in the order specified by the given priorities *)
+let%client prioritized_waiters : (int option * unit Lwt.t * bool) list ref =
+  ref []
+
+(* [Lwt.cancel] does nothing if the task is already resolved,
+   so we can safely cancel them all *)
+let%client cancel_waiters () =
+  List.iter (fun (_,w,_) -> Lwt.cancel w) !prioritized_waiters
+
+(* This boolean is used to track whether
+   the list of priorities has been sorted
+ *)
+let%client sorted = ref false
+
+(* A priority of [None] is considered infinite,
+   and thus greater than everything else *)
+let%client compare_priority_opt p1 p2 =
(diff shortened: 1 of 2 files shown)


src/os_tips.eliomi
```diff
@@ -40,6 +40,10 @@
     for connected users only, non-connected users only, or all (default).
     Tips for non-connected users will reappear every time the session is closed.
     - [?delay] adds a delay before displaying the tip (in seconds)
+    - [?priority] specifies a lower-value-first priority order for this bubble.
+    Priority 1 bubbles will be displayed first, Priority [None] bubbles will
```
balat, line 44: And priority -237?

src/os_tips.eliom
```diff
@@ -189,12 +189,101 @@ let%shared block ?(a = []) ?(recipient = `All)
...
+  | _ ->
+    (* Catch-all because everything else uses guards,
+       but should be unreachable *)
+    assert false
+
+let%client wait_for_bubble ?priority () =
+  (* We wait for the elements to load,
+     to be sure we have all waiters prioritized *)
+  let%lwt () = onload_waiter () in
+  if not !sorted then
```
balat, line 265: ocamlformat failed. We should probably add a commit hook?

src/os_tips.eliom
```diff
@@ -189,12 +189,101 @@ let%shared block ?(a = []) ?(recipient = `All)
...
+    (prioritized_waiters := List.rev !prioritized_waiters; sorted := true);
```
balat, line 266: Why not sorting in reverse order directly?

src/os_tips.eliom
clembu: failed? Do we have a `.ocamlformat` in this repo?

src/os_tips.eliom
clembu: It would reverse the call-order too. On priority equality, first calls are inserted first, thus later in the list. We have no other reliable way to keep this order than reversing the list.

src/os_tips.eliom
balat: Damned. Only in the template 😭

src/os_tips.eliom
balat: ok thanks

clembu: That said, I'm currently sorting the list on every addition, which is suboptimal. I'll move the sort to this place so we only do it once (per batch)

PR #636, opened 2021-11-03
compatibility of template distillery with outsourced ocsipersist

jrochel:
Depends on: https://github.com/ocsigen/ocsigen-start/pull/635


jrochel:
@Zzull ping

PR #641, opened 2021-12-16
activation links: add short grace period before expiration

vouillon:
It seems the activation link is sometimes opened several times (once in a navigator and once in the app?) on some devices. So that it can still be use in this case, we do not expire it once away but let it active for 20 seconds.

### DIFF
--- a/src/os_db.ml
+++ b/src/os_db.ml
@@ -1531,70 +1531,136 @@ module User =
                          let v = max 0L (Int64.pred validity) in
                          [%lwt
                            let () =
-                             PGOCaml.bind
-                               (let dbh = dbh in
-                                let params : string option list list =
-                                  [[Some
-                                      (((let open PGOCaml in string_of_int64))
-                                         v)];
-                                  [Some
-                                     (((let open PGOCaml in string_of_string))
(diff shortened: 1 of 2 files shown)


src/os_db.ml
```diff
@@ -1531,70 +1531,136 @@ module User =
...
+                                       PGOCaml.bind
+                                         (PGOCaml.prepare dbh ~name ~query ())
+                                         (fun () ->
+                                            Hashtbl.add hash name true;
+                                            PGOCaml.return ())
+                                     else PGOCaml.return ())
+                                    (fun () ->
+                                       PGOCaml.execute_rev dbh ~name ~params
+                                         ()))
+                                 (fun _rows -> PGOCaml.return ()) in
```
jrochel, line 1663: Is it particularly difficult to share code between the `then` and the `else` branch? I'm having trouble finding the differences and the commonalities between them.

jrochel (approved):

src/os_db.ml
vouillon: This code is generated by the PGOCaml ppx from `os_db.ppx.ml`

PR #645, opened 2022-02-11
[actions] fix ssl error during Github Actions compilation on macos

jrochel:
opam depext doesn't seem to have an effect on MacOS.
It only prints the packages one needs to install manually.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,8 +52,8 @@ jobs:
       - run: opam install -y .
         working-directory: template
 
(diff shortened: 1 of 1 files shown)


Zzull:
@jrochel

Zzull:
Almost there…

jrochel:
> Almost there…

Actually I'd say that this PR can be merged.
The remaining compilation errors on MacOS occur non-deterministically.

ilankri (approved):

PR #683, opened 2024-09-23
Fix SASS generation in Makefile.style

grazil:
Hello,

I detect an error in Makefile.style that blocked sass generation.


balat:
Thanks!

PR #687, opened 2025-03-05
Remove `lwt_ppx` dependency

Julow:
This PR removes the use of `lwt_ppx` by rewriting `let%lwt`, `match%lwt`, etc.. to `let*`, `Lwt.bind` and other direct calls to `Lwt` functions.
Templates are also updated.

This was done using this tool: https://github.com/Julow/lwt_ppx_to_let_syntax/

The first commit runs OCamlformat to remove parasitic diffs in the other commits as the rewrite tool relies on OCamlformat too.

This is a first step toward moving from Lwt to effect-based concurrency. Here are related PRs on the other ocsigen projects: https://github.com/ocsigen/eliom/pull/815 https://github.com/ocsigen/ocsigen-toolkit/pull/236 https://github.com/ocsigen/ocsipersist/pull/6


src/os_db.ml
Julow, line 1: I wasn't sure how to update this file so I let the tool run on it. Any guideline on how to setup the right environment to run the update command ?

src/os_db.ml
```diff
@@ -14,8 +16,8 @@ let one f ~success ~fail q = f q >>= function r :: _ -> success r | _ -> fail
 
 let pwd_crypt_ref =
   ref
-    ( (fun password -> Bcrypt.string_of_hash (Bcrypt.hash password))
-    , fun _ password1 password2 ->
+    ( (fun password -> Bcrypt.string_of_hash (Bcrypt.hash password)),
+      fun _ password1 password2 ->
```
balat, line 20: Did you ocamlformat with our .ocamlformat?

balat (approved):

src/os_db.ml
balat: Maybe. In that case we should probably remove it from the repository, isn't it?

Julow: `os_db.ml` is uptodate ! This is ready to be merged.

PR #706, opened 2026-06-08
Declare ocsigen-dune-rules as a build dependency

balat:
The build runs the ocsigen-dune-rules generator but never declared it; a clean opam install failed with 'Program ocsigen-dune-rules not found'.


Julow (commented):
This doesn't make sense. Ocsigen-dune-rules is vendored

balat:
I probably chose the wrong base branch for the PR. It is supposed to fix an issue with recent versions where it is not vendored any more. I'll fix that

PR #725, opened 2026-06-30
ci(doc): cache the opam switch to skip recompiling the dependency tree

balat:
## Problem

The `Documentation` workflow recompiles the whole dependency tree (incl. wasm_of_ocaml) plus wodoc and odoc from source on **every push** (several minutes); `setup-ocaml@v3` only caches the base switch (the compiler).

## Fix

Cache the whole `_opam` switch with `actions/cache`, keyed on `*.opam` (the dependency set) and wodoc's resolved HEAD commit. The install is split:

- **Install dependencies and wodoc** — cache miss only (deps tree + wodoc).
- **Install <project>** — **always**: the project's own sources change every push, so `opam install .` re-syncs the local pin and rebuilds it for `odoc_driver --remap` to document. Only the project recompiles on the hot path, not its dependencies.

The cache rebuilds once when an `*.opam` file or wodoc's HEAD move.

Same approach validated on ocsigen.github.io and ocsigenserver (each ~5 min → ~1 min).

> ⚠️ This project pulls wasm_of_ocaml; if a validation run hits the current `Aandreba/setup-binaryen`/wasm_of_ocaml build breakage (see ocsigen/eliom#894), that is **independent of this change** and would block the next master push too.
> The `release` job (manual, infrequent) is left as-is.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -50,20 +50,42 @@ jobs:
         with:
           token: ${{ github.token }}
 
-      - name: Install dependencies
+      # The bottleneck of this job is compiling the dependency tree (incl.
+      # wasm_of_ocaml) plus wodoc and odoc from source; setup-ocaml only caches the
+      # base switch (the compiler), so they recompiled on every push. Cache the
+      # whole _opam switch keyed on ocsigen-start.opam (the dep set) and wodoc's HEAD
+      # commit. On a cache hit the deps and wodoc are restored and only ocsigen-start
+      # itself is (re)built below; the key changes -- and the deps/wodoc rebuild
+      # once -- only when ocsigen-start.opam or wodoc move.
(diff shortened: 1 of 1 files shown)


balat:
⚠️ The validation dispatch on this branch **failed at `Install dependencies and wodoc`, but not because of this change** — it died building `wasm_of_ocaml-compiler 6.4.0` (pulled in via eliom), the same `Aandreba/setup-binaryen`/wasm_of_ocaml breakage as ocsigen/eliom#894 and eliom. This is **independent of caching** and would block the next master doc push too.

So this caching PR is logically ready but **can't be validated until the wasm build is fixed**. The pattern itself (cache deps+wodoc, always rebuild the project via `opam install .`) is validated on the equivalent API projects (ocsigenserver: 5 min → ~1 min) and on the dogfood/manual variants (wodoc, ocsigen.github.io).

